PR #85, opened 2025-02-06
feat(aws): introduce AWS S3 + Lambda@Edge infra 

gronxb:
WIP

### DIFF
--- a/packages/hot-updater/package.json
+++ b/packages/hot-updater/package.json
@@ -72,6 +72,7 @@
     "@types/node": "^22.9.0",
     "@types/plist": "^3.0.5",
     "@types/semver": "^7.5.8",
+    "aws-sdk": "^2.1692.0",
     "boxen": "^8.0.1",
     "dayjs": "^1.11.13",
     "execa": "^9.5.2",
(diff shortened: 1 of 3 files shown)


plugins/aws/tsup.config.ts
```diff
@@ -1,24 +1,11 @@
...
-      js: `import { createRequire } from 'module'; const require = createRequire(import.meta.url);`,
-    },
+export default defineConfig({
+  entry: ["src/index.ts", "sdk/index.ts", "lambda/index.ts"],
+  format: ["esm", "cjs"],
+  outDir: "dist",
+  dts: true,
+  banner: {
+    js: `import { createRequire } from 'module'; const require = createRequire(import.meta.url);`,
   },
```
rjwu95, line 10: @gronxb 
I’m not sure why, but when I upload a zip file with this configuration, the Lambda doesn’t execute properly.

rjwu95: but noExternal execute well.

gronxb: I think it should work the same way since moving a dependency from `dependencies` to `devDependencies` makes tsup automatically treat it as noExternal. But does it still not work?

rjwu95: Lambda@Edge can only run in a CJS (CommonJS) environment, but if an index.js file exists, Lambda@Edge recognizes it as the default entry file.
The index.cjs file is executable since it follows CJS, but it seems that the banner added in tsup.config.ts is causing an error when executing it in CJS due to the import statement.

so I opened a new [PR](https://github.com/gronxb/hot-updater/pull/121).

PR #88, opened 2025-02-07
fix(react-native): `runUpdateProcess` RunUpdateProcessResponse type optional

qnrjs42:
Before Fix)
스크린샷 2025-02-07 오후 12 17 30

After Fix)
스크린샷 2025-02-07 오후 12 18 26

### DIFF
--- a/packages/react-native/src/runUpdateProcess.ts
+++ b/packages/react-native/src/runUpdateProcess.ts
@@ -3,14 +3,10 @@ import { reload, updateBundle } from "./native";
 
 export type RunUpdateProcessResponse =
   | {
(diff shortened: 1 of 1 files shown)


gronxb:
Thank you for your contribution! However, this is not a bug but an intended behavior.

When the status is `UP_TO_DATE`, only the status exists. The type is properly exposed once the status is confirmed as `UPDATE` or `ROLLBACK`.

image

image

qnrjs42:
Thank you for the quick feedback!

I only looked at a part of the code when making the modification, so I didn’t realize it was the intended behavior.
I hope the documentation gets updated.

I will close the Pull Request.

PR #99, opened 2025-02-14
feat(react-native): add `status` to fallbackComponent in wrap function

purrsong-chong:
**Description**
In this PR, I have added status as a prop to the line 
`return <Fallback progress={progress} status={status} />; `
in the wrap.tsx file.

The purpose of adding the status prop is to allow the fallbackComponent to provide more dynamic and informative UI feedback based on the current update process.

### DIFF
--- a/docs/docs/guide/hot-updater/wrap.mdx
+++ b/docs/docs/guide/hot-updater/wrap.mdx
@@ -40,7 +40,7 @@ The `fallbackComponent` receives two props: `progress` and `shouldForceUpdate`.
 
 - **progress**: This prop indicates the download progress of the new bundle. You can use this to display a progress bar or percentage to the user.
   
-- **status**: This prop indicates the current status of the update process. It can be one of the following: `"IDLE"`, `"CHECK_FOR_UPDATE"`, `"UPDATING"`, or `"UPDATE_PROCESS_COMPLETED"`. You can use this to display different messages or UI elements based on the update status.
+- **status**: This prop indicates the current status of the update process. It can be one of the following: `CHECK_FOR_UPDATE`, `UPDATING`. You can use this to display different messages or UI elements based on the update status.
 
 By using these props, you can customize the user interface to provide feedback during the update process, ensuring a smooth transition to the updated version of your app.
 
@@ -58,7 +58,7 @@ function App() {
 
 export default HotUpdater.wrap({ // [!code hl:22]
   source: "<your-update-server-url>",
-  fallbackComponent: ({ progress = 0, status }) => (
+  fallbackComponent: ({ progress, status }) => (
       <View
         style={{
           flex: 1,
@@ -69,8 +69,10 @@ export default HotUpdater.wrap({ // [!code hl:22]
           backgroundColor: "rgba(0, 0, 0, 0.5)",
         }}
       >
+        {/* You can put a splash image here. */}
+
         <Text style={{ color: "white", fontSize: 20, fontWeight: "bold" }}>
-          {status === "UPDATING" ? "Updating..." : "Preparing Update..."}
+          {status === "UPDATING" ? "Updating..." : "Checking for Update..."}
         </Text>
         <Text style={{ color: "white", fontSize: 20, fontWeight: "bold" }}>
           {Math.round(progress * 100)}%
--- a/docs/docs/guide/hot-updater/wrap.mdx
+++ b/docs/docs/guide/hot-updater/wrap.mdx
@@ -36,6 +36,15 @@ This allows you to show the progress while downloading the new bundle and safely
 
 If you don't define a `fallbackComponent`, the bundle will be downloaded without blocking the screen.
 
+The `fallbackComponent` receives two props: `progress` and `shouldForceUpdate`.
+
+- **progress**: This prop indicates the download progress of the new bundle. You can use this to display a progress bar or percentage to the user.
+  
+- **shouldForceUpdate**: This boolean prop indicates whether a force update is required. If `true`, it means the update is critical, and you should ensure the user cannot bypass the update process.
+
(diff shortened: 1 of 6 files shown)


packages/react-native/src/wrap.tsx
```diff
@@ -29,7 +29,11 @@ export interface HotUpdaterConfig extends CheckForUpdateConfig {
    *
    * If not defined, the bundle will download in the background without blocking the screen.
    */
-  fallbackComponent?: React.FC<Pick<HotUpdaterState, "progress">>;
+  fallbackComponent?: React.FC<
+    Pick<HotUpdaterState, "progress"> & {
+      shouldForceUpdate: boolean;
+    }
+  >;
```
gronxb, line 36:
```suggestion
  fallbackComponent?: React.FC<{
    progress: number;
    shouldForceUpdate: boolean;
  }>;
```

packages/react-native/src/wrap.tsx
```diff
@@ -73,6 +78,8 @@ export function wrap<P>(
             return;
           }
 
+          setShouldForceUpdate(updateInfo.shouldForceUpdate);
+
```
gronxb, line 82: Is state really necessary? Wouldn’t it be enough to pass the variable directly as a prop to the Fallback component? The value won’t change at the time it is passed as a prop.

gronxb (changes requested):
Thank you for your contribution!

packages/react-native/src/wrap.tsx
purrsong-chong: Yes, you are right

gronxb (changes requested):
I made a mistake. I originally thought the entire codebase followed a single flow, but I was wrong.

Now that I see it, managing it as a state makes more sense. Sorry about that—please revert it to the original approach.

Also, could you update the documentation to reflect this? The file is located at docs/docs/guide/hot-updater/wrap.mdx.

purrsong-chong:
> I made a mistake. I originally thought the entire codebase followed a single flow, but I was wrong.
> 
> Now that I see it, managing it as a state makes more sense. Sorry about that—please revert it to the original approach.
> 
> Also, could you update the documentation to reflect this? The file is located at docs/docs/guide/hot-updater/wrap.mdx.
> 
> https://gronxb.github.io/hot-updater/guide/hot-updater/wrap.html#fallback-component

I have reverted the changes to manage the state as it makes more sense given the codebase's structure. 
Additionally, I have updated the documentation in docs/docs/guide/hot-updater/wrap.mdx to reflect this change. 
The status prop has also been added to the fallbackComponent to provide more detailed feedback during the update process.

docs/docs/guide/hot-updater/wrap.mdx
```diff
@@ -51,25 +60,28 @@ function App() {
...
-      <Text style={{ color: "white", fontSize: 20, fontWeight: "bold" }}>
-        Updating...
-      </Text>
-      <Text style={{ color: "white", fontSize: 20, fontWeight: "bold" }}>
-        {Math.round(progress * 100)}%
-      </Text>
-    </View>
-  ),
+  fallbackComponent: ({ progress = 0, shouldForceUpdate }) => {
+    if(!shouldForceUpdate) return null;
```
gronxb, line 64: As long as the fallback component prop exists, returning null will not be ignored.

This is an intentional behavior to block app entry during the update check.

Therefore, returning null in the example wouldn’t have much significance. The key point is that a flag can be used during the fallback process.

gronxb (changes requested):

docs/docs/guide/hot-updater/wrap.mdx
gronxb: and status prop exports

purrsong-chong: I got it. I should take a different approach.
Thank you for your feedback.

gronxb: Separately, the approach of passing status and shouldForceUpdate to the fallback component to provide more information seems like a good idea, so I think it would be worth reopening the PR.

gronxb: In the fallback component, some users thought it was a bug because the progress remained at 0 while checking for updates. Passing a status value would make it explicitly clear that an update check is in progress, allowing the UI to reflect that accordingly. Therefore, in the example code, removing the return null when shouldForceUpdate is true seems like a reasonable approach.

gronxb: What I intended was to temporarily block the app’s entry point and display a splash image when using a fallback component.

gronxb: **However, since ignoring the fallback when shouldForceUpdate is false seems like a reasonable option, I’ll consider a different approach myself.**

purrsong-chong: I think that in the current flow, the presence or absence of `shouldForceUpdate` does not change the download process. Also, as you pointed out, blocking the UI in the `fallbackComponent` based on `shouldForceUpdate` seems to be meaningless. It would be better to remove `shouldForceUpdate` and only pass `status` as a prop.

gronxb (approved):
I did a refactoring! I’ll go ahead and merge it. Thanks for your contribution!

gronxb:
I considered that when shouldForceUpdate is false and an update is available, it qualifies as a minor update. Therefore, it is correct for the fallbackComponent not to be triggered. I determined this as a bug and fixed it.

After re-reading the documentation, I gained a clearer understanding of when blocking occurs.

This fix is scheduled for release in version 0.6.7.

gronxb:
https://github.com/gronxb/hot-updater/releases/tag/v0.6.7

PR #111, opened 2025-02-17
fix(ios): ensure UI thread is not blocked by sending events

ewindso:
…g out by 200ms


gronxb (approved):

gronxb:
Is there a test result (video) available? If you share the result, we can merge it more quickly.

gronxb:
Release in 0.9.0

ewindso:
Thanks!  @gronxb I would show it but would need permission

PR #113, opened 2025-02-17
fix: bundle Lambda@Edge dependencies into a single file

rjwu95:
- minor Lambda fix
 - remove mistakenly committed file


plugins/aws/lambda/index.ts
```diff
@@ -70,7 +67,7 @@ export const handler: CloudFrontRequestHandler = async (event) => {
 
   const bundles = results
     .filter((result) => result.status === "fulfilled")
-    .map((result) => result.value) as Bundle[];
+    .map((result) => result.value).flat() as Bundle[];
```
gronxb, line 70:
```suggestion
    .flatMap((result) => result.value) as Bundle[];
```

gronxb (changes requested):
Run the following command:

`pnpm -w biome`

PR #115, opened 2025-02-18
Ensuring there's no UI lock on Android, also preventing crash

ewindso:
hi @gronxb , I can't provide a before/after video of the app as it's private, but this is what happened:

Before: Stuck on app launch when there's an update, eventually it might work / crash

After: App loads update, no problem.

### DIFF
--- a/packages/react-native/android/src/oldarch/HotUpdaterModule.kt
+++ b/packages/react-native/android/src/oldarch/HotUpdaterModule.kt
@@ -1,10 +1,14 @@
 package com.hotupdater
 
-import com.facebook.react.bridge.Promise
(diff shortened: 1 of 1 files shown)


gronxb:
It seems that the old architecture failed in CI because of Java. Will the modified code be able to run in Java?

ewindso:
Ohhh @gronxb good point...  hmm yeah this will only work with Kotlin right now...

Gautham495:
I am facing the same issue @ewindso

gronxb:
This issue resolved #122. I'll close it, thank you!

gronxb:
Released In 0.10.1

ewindso:
Thanks @gronxb !

PR #119, opened 2025-02-18
fix(ios): Update bundleUrl Before Reload for Custom RCTBridges

wes4m:
# Issue
We have a brownfield app, meaning it originally started as a native app, and React Native was integrated later. As a result:
- The entry point is not React Native.
- We manually create the RCTBridge.
- Our AppDelegate does not conform to RCTAppDelegate.

During a forced OTA update (when reload = true), hot-updater triggers `RCTTriggerReloadCommandListeners` this leads to the following issue:
- `RCTTriggerReloadCommandListeners` reaches the registered bridge, which still points to the old bundle URL.
- App reloads the last set bundleUrl on that bridge, which is not the new one.
- The old bundle is loaded again triggering an update check and another reload.
- This creates an infinite reload loop, despite the new bundle being downloaded correctly.
Note that If reload is not triggered, the new bundle loads correctly on the second run without any issues which makes sense.

This scenario is handled by CodePush by updating the super bridge's bundle URL before triggering the reload. This PR does the same thing. 

I believe this change should have no impact on other typical React Native setups.

### DIFF
--- a/packages/react-native/ios/HotUpdater/HotUpdater.mm
+++ b/packages/react-native/ios/HotUpdater/HotUpdater.mm
@@ -252,6 +252,7 @@ RCT_EXPORT_MODULE();
 RCT_EXPORT_METHOD(reload) {
     NSLog(@"HotUpdater requested a reload");
     dispatch_async(dispatch_get_main_queue(), ^{
+        [super.bridge setValue:[HotUpdater bundleURL] forKey:@"bundleURL"];
         RCTTriggerReloadCommandListeners(@"HotUpdater requested a reload");
     });
 }


gronxb:
If the bundle URL is set during a reload, what happens in cases where it’s not a forced update? Since reload is not triggered when it’s not a forced update, the system applies the update when the app is restarted. Will it still work correctly in that case?

wes4m:
> Thanks your contribution.
> 
> If the bundle URL is set during a reload, what happens in cases where it’s not a forced update? Since reload is not triggered when it’s not a forced update, the system applies the update when the app is restarted. Will it still work correctly in that case?

Yes when it's a regular update there are no issues, as the RCTBridge will be initialized again on app start using `HotUpdater.bundleURL()` which will always return the updated bundle url.

gronxb (approved):
Thanks !!

PR #133, opened 2025-02-21
feat(console): Improve UX by preventing screen flickering

jingjing2222:
1. Splash screen displayed only on initial application entry.
2. Minimized data fetching frequency and eliminated visual flickering using @tanstack/solid-query's query key feature.
* (Note: solid-router version 0.14.10 prevents the direct usage of query in route components. Data prefetching or alternative strategies may be needed).

https://github.com/user-attachments/assets/4d097895-2a94-4a5f-bae1-95ae382cf779

Closed #131

### DIFF
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -35,6 +35,7 @@
     "@solidjs/meta": "^0.29.4",
     "@solidjs/router": "^0.14.10",
     "@tanstack/solid-form": "^0.34.0",
+    "@tanstack/solid-query": "^5.66.4",
     "@tanstack/solid-table": "^8.20.5",
     "@tanstack/valibot-form-adapter": "^0.34.0",
     "@types/node": "^22.8.6",
@@ -57,8 +58,7 @@
     "valibot": "^0.42.1"
   },
   "dependencies": {
-    "@hot-updater/core": "0.10.1",
-    "@hot-updater/plugin-core": "0.10.1",
-    "@tanstack/solid-query": "^5.66.4"
+    "@hot-updater/core": "0.11.0",
+    "@hot-updater/plugin-core": "0.11.0"
   }
 }
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -57,7 +57,8 @@
     "valibot": "^0.42.1"
   },
   "dependencies": {
-    "@hot-updater/core": "0.11.0",
-    "@hot-updater/plugin-core": "0.11.0"
+    "@hot-updater/core": "0.10.1",
+    "@hot-updater/plugin-core": "0.10.1",
+    "@tanstack/solid-query": "^5.66.4"
   }
 }
--- a/packages/console/src/App.tsx
+++ b/packages/console/src/App.tsx
@@ -1,23 +1,29 @@
 import { MetaProvider, Title } from "@solidjs/meta";
 import { Route, Router } from "@solidjs/router";
-import { Suspense } from "solid-js";
 import "./App.css";
-import { Toaster } from "./components/ui/sonner";
+import { SplashScreen } from "@/components/spash-screen";
+import Layout from "@/components/ui/layout";
+import { QueryClient, QueryClientProvider } from "@tanstack/solid-query";
+import { Suspense } from "solid-js";
 import Home from "./routes";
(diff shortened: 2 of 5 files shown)


packages/console/package.json
```diff
@@ -57,7 +57,8 @@
     "valibot": "^0.42.1"
   },
   "dependencies": {
-    "@hot-updater/core": "0.11.0",
-    "@hot-updater/plugin-core": "0.11.0"
+    "@hot-updater/core": "0.10.1",
+    "@hot-updater/plugin-core": "0.10.1",
+    "@tanstack/solid-query": "^5.66.4"
```
gronxb, line 62: move to devDeps

packages/console/src/App.tsx
```diff
@@ -1,23 +1,42 @@
...
+import { Show, Suspense, createSignal, onMount } from "solid-js";
 import Home from "./routes";
+export const queryClient = new QueryClient();
 
 export default function App() {
+  const [onSuspense, setOnSuspense] = createSignal(false);
+
+  onMount(() => {
+    setTimeout(() => {
+      setOnSuspense(true);
```
gronxb, line 16: why  need this?

gronxb (changes requested):

packages/console/src/App.tsx
jingjing2222: I needed that before install @tanstack-query, but now doesn't.
I will modify that logic soon.

jingjing2222 (commented):
I did you said, Check plz

packages/console/src/App.tsx
```diff
@@ -1,23 +1,29 @@
...
 import { Route, Router } from "@solidjs/router";
-import { Suspense } from "solid-js";
 import "./App.css";
-import { Toaster } from "./components/ui/sonner";
+import { SplashScreen } from "@/components/spash-screen";
+import Layout from "@/components/ui/layout";
+import { QueryClient, QueryClientProvider } from "@tanstack/solid-query";
+import { Suspense } from "solid-js";
 import Home from "./routes";
+export const queryClient = new QueryClient();
```
gronxb, line 9: Since it is being provided through a provider, direct access from other places must be prevented.
```suggestion
const queryClient = new QueryClient();
```

packages/console/src/routes/index.tsx
```diff
@@ -42,52 +42,45 @@ export default function Home() {
...
-          onRowClick={(row) => {
-            start(() => {
-              setSelectedBundleId(row.id);
-            });
-          }}
-        />
+  const handleClose = () => {
+    start(() => {
+      setSelectedBundleId(null);
+      queryClient.invalidateQueries({ queryKey: ["getBundles"] });
```
gronxb, line 50: Isn’t it okay if invalidate doesn’t happen when closing?

gronxb (changes requested):

packages/console/src/routes/index.tsx
```diff
@@ -21,12 +17,16 @@ import { EditBundleSheetContent } from "./_components/edit-bundle-sheet-content"
...
   const params = useParams();
   const navigate = useNavigate();
+  const queryClient = useQueryClient();
 
   const bundleId = params.bundleId;
 
-  const [data, { refetch }] = createResource(() =>
-    api.getBundles.$get().then((res) => res.json()),
-  );
+  const data = createQuery(() => ({
```
gronxb, line 24: It might be a good idea to extract it into hooks:
	•	useBundles() → returns all bundles
	•	useBundle(id) → returns a specific bundle

gronxb (changes requested):

packages/console/src/routes/index.tsx
jingjing2222: Console already has that getBundelById(id) in `EditBundleSheetContent` Component
I think getBundles is better in index.tsx

packages/console/src/routes/index.tsx
jingjing2222: `handleClose` function is only activated with Clicking save button.
I think improving invalidation is safer.. but I don't think that has to be.
How about you? You don't think that as me?

gronxb (approved):

PR #137, opened 2025-02-22
feat(console): improve missing sidebar with animation

jingjing2222:
Before this PR, sidebar doesn't show animation during closed.
I added delay while it works.
When click outside of sidebar, move `'/'` also

### DIFF
--- a/packages/console/src/lib/utils.ts
+++ b/packages/console/src/lib/utils.ts
@@ -4,3 +4,5 @@ import { twMerge } from "tailwind-merge";
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs));
 }
+export const sleep = (second: number) =>
+  new Promise((resolve) => setTimeout(resolve, second));
--- a/packages/console/src/routes/index.tsx
+++ b/packages/console/src/routes/index.tsx
@@ -1,14 +1,9 @@
 import { Sheet } from "@/components/ui/sheet";
 import { createBundlesQuery } from "@/lib/api";
+import { sleep } from "@/lib/utils";
 import { useNavigate, useParams } from "@solidjs/router";
-import { createMemo } from "solid-js";
-import {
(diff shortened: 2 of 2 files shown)


packages/console/src/lib/utils.ts
```diff
@@ -4,3 +4,5 @@ import { twMerge } from "tailwind-merge";
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs));
 }
+export const sleep = (second: number) =>
+  new Promise((resolve) => setTimeout(resolve, second));
```
gronxb, line 8:
```suggestion

export const sleep = (ms: number) =>
  new Promise((resolve) => setTimeout(resolve, ms));
```

packages/console/src/routes/index.tsx
```diff
@@ -35,35 +28,28 @@ export default function Home() {
     navigate(`/${selectedBundleId()}`, { replace: true });
   });
 
+  createEffect(() => {
+    if (open()) return;
```
gronxb, line 32:
```suggestion
    if (open()) {
        return;
    }
```

gronxb (changes requested):

packages/console/src/routes/index.tsx
```diff
@@ -35,35 +28,28 @@ export default function Home() {
...
+    sleep(500).then(() => setSelectedBundleId(null));
+  });
+
   const dataForTable = createMemo(() => data.data || []);
+  const [open, setOpen] = createSignal(true);
```
gronxb, line 37:
```suggestion
  const [isOpen, setIsOpen] = createSignal(true);
```

gronxb (changes requested):

jingjing2222:
Apply your comment to commit, Check plz 👍

gronxb (approved):

PR #142, opened 2025-02-24
platform specific config file location

mstfmedeni:
There is no platform-based control in config file control. Since it is called with console and run with nodejs, it is not possible to use platform in ts. Therefore, after checking the platform with console, a platform-specific config file path list was added.

Why is it necessary? 
we use hermes for ios in our project but not for android. we need to check and change hermes on platform basis.


gronxb:
https://github.com/gronxb/hot-updater/pull/143

Thank you for the report. This approach seems to be a simpler solution!

gronxb:
https://github.com/gronxb/hot-updater/releases/tag/v0.12.3

Issue #110, opened 2025-02-16
Feature Request: Firebase Integration as Alternative Backend Provider


Subramanyarao11:
## Overview
First, thank you for creating this valuable solution for OTA updates as an alternative to CodePush. I'm currently exploring hot-updater for our project and have a feature request/question regarding backend provider flexibility.

## Current Setup
Our application heavily utilizes Firebase services including:
- Push Notifications ( FCM )
- Crashlytics
- Remote Config

## Feature Request
Would it be possible to add support for Firebase as an alternative backend provider instead of Supabase? Specifically:
- Using Firebase Storage for bundle storage
- Leveraging Firebase Cloud Functions for the edge function implementation

## Benefits
- Reduced infrastructure complexity for teams already using Firebase
- Single provider for multiple app services
- Consolidated monitoring and billing
- Familiar tooling and SDK

## Questions
1. Is this something that aligns with the project's vision?
2. Are there any technical limitations that would make this integration challenging?
3. Would you be open to contributions implementing this feature?

I'm happy to provide more details or clarification if needed.

gronxb:
> 1. Is this something that aligns with the project's vision?

Yes, it does. I believe supporting Firebase would be beneficial.

> 2. Are there any technical limitations that would make this integration challenging?

I think the biggest challenge would be providing an init command that deploys to Firebase Cloud Functions. `hot-updater` helps users get up and running quickly by handling infrastructure tasks. For example, in the Supabase case, when running `hot-updater init`, the CLI creates a bucket, performs database table migrations, and finally helps deploy the edge function. If we decide to support Firebase, these steps need to be streamlined to ensure a good developer experience.

> 3. Would you be open to contributions implementing this feature?

At the moment, I have other higher-priority tasks, so this isn’t at the top of my list. However, I’m definitely open to contributions. If someone were to submit a contribution, I’d be more than happy to accelerate the process.

gronxb:
Below is the guide for anyone who wants to contribute a Firebase integration. The plan is to add two new plugins under [plugins/firebase](https://github.com/gronxb/hot-updater/tree/main/plugins):
1. `firebaseStorage`
2. `firebaseDatabase`

## 1. `firebaseStorage`
Reference:
[StoragePlugin definition](https://github.com/gronxb/hot-updater/blob/8bdc1bd683fed9489a1491beb550f6b4a5490fb1/plugins/plugin-core/src/types.ts#L38)

```ts
export interface StoragePlugin {
[...]

PR #149, opened 2025-02-25
feat(firebase): integration `firebaseStorage`

jingjing2222:
issue #110 
I did the first step of firebase integration. I added firebaseStorage.

in firebase

1. make project
2. make webapp project
3. there are keys for environment variables
image

### DIFF
--- a/plugins/firebase/package.json
+++ b/plugins/firebase/package.json
@@ -1,0 +1,42 @@
+{
+  "name": "@hot-updater/firebase",
+  "private": true,
+  "type": "module",
+  "version": "0.12.3",
+  "description": "React Native OTA solution for self-hosted",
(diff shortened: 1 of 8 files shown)


plugins/firebase/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@hot-updater/firebase",
+  "private": true,
```
gronxb, line 3:
```suggestion
```

gronxb (changes requested):

jingjing2222:
https://github.com/user-attachments/assets/8b219b22-7325-43e2-aefd-5fc5add9d16c

After adding init, it must have security pattern, but I think this hot-updater is working on personal console.

It seems to be afterthought.

PR #156, opened 2025-02-27
feat(firebase): integration `firebaseDatabase`

jingjing2222:
first, I add firebaseDatabase, It is only for `existing` `firestore Database`
second, I add test code for Database, I think It will be upgraded after add init setting.

So there are some works for security, testing upgrade, and so on.

https://github.com/user-attachments/assets/e028d690-8990-4f64-b37c-1382b1597c59

#110


gronxb (approved):

plugins/firebase/src/firebaseDatabase.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export interface FirebaseDatabaseConfig {
+  apiKey: string;
+  projectId: string;
+  authDomain?: string;
+  storageBucket?: string;
+  messagingSenderId?: string;
+  appId?: string;
+  appName?: string;
+}
```
gronxb, line 28: reduce config

PR #161, opened 2025-02-28
chore(cli): Skip sourcemap files when creating bundle zip

rjwu95:


packages/hot-updater/src/utils/createZip.ts
```diff
@@ -18,6 +18,11 @@ export const createZip = async ({
     files.sort();
 
     for (const file of files) {
+      if (file.endsWith(".map")) {
+        // skip sourcemap files
+        continue;
+      }
```
gronxb, line 24: Since createZip is a utility that only compresses zip files, how about controlling its behavior using options?

```js
createZip({
   outfile: "..",
   targetDir: "..",
   excludeExts: [".map"]
});
```

packages/hot-updater/package.json
gronxb, line 1: please revert

gronxb (changes requested):

packages/hot-updater/src/utils/createZip.ts
rjwu95: done!

packages/hot-updater/package.json
rjwu95: done!

packages/hot-updater/src/commands/init/aws/index.ts
```diff
@@ -220,6 +220,7 @@ export const deployLambdaEdge = async (
           await createZip({
             outfile: zipFilePath,
             targetDir: lambdaDir,
+            excludeExts: [".map"],
```
gronxb, line 223:
```suggestion
```

gronxb (changes requested):

gronxb (approved):
Thanks !

Issue #150, opened 2025-02-25
Rollbacks on android to original bundle (000) resulting in a soft crash


regalstreak:
The app starts if we reopen it for a 2nd time

```
/libnfc_nci: nfaConnectionCallback: NFA_PRESENCE_CHECK_EVT
D/libnfc_nci: nativeNfcTag_doPresenceCheck(1349): isPresent = 1
E/unknown:com.facebook.react.uimanager.UIManagerHelper: Unhandled SoftException
E/unknown:com.facebook.react.uimanager.UIManagerHelper: com.facebook.react.bridge.ReactNoCrashSoftException: Cannot get UIManager because the context doesn't contain an active CatalystInstance.
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at com.facebook.react.uimanager.UIManagerHelper.getUIManager(UIManagerHelper.java:82)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at com.facebook.react.uimanager.UIManagerHelper.getUIManager(UIManagerHelper.java:52)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at com.facebook.react.uimanager.UIManagerModule.dispatchViewManagerCommand(UIManagerModule.java:577)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at com.facebook.react.uimanager.ReanimatedUIManager.dispatchViewManagerCommand(ReanimatedUIManager.java:170)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at java.lang.reflect.Method.invoke(Native Method)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:146)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at com.facebook.jni.NativeRunnable.run(Native Method)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at android.os.Handler.handleCallback(Handler.java:991)
E/unknown:com.facebook.react.uimanager.UIManagerHelper: 	at android.os.Handler.dispatchMessage(Handler.java:102)
[...]

PR #168, opened 2025-03-07
release: 0.13.0

### LINKED ISSUE #86: How to use multiple mobile apps in the same project on Supabase with different buckets?
I have decided to use Supabase as the provider, but I am facing a problem. I have about 10 mobile apps, and I only want to create a single project on Supabase to push hotfixes for all 10 mobile apps. However, I'm encountering an issue where the updates are applied to all of my apps. What I want is for each mobile app to only retrieve updates from the corresponding buckets created for each app at initialization. I would really like to know if this is possible or if I must create a separate project on Supabase for each app. Thank you very much!
@gronxb

### LINKED ISSUE #155: App is refreshing again once i downloaded the app from Test Flight
I have added the hot-updater using the supabase provider and it's working fine. but there was only one issue when I installed the app from the Test Flight every time the app refreshed after installing the app from test flight because in the last bundle update, I deployed the bundle with the option. 

yarn hot-updater deploy -i -f

Expected Result: The app should download the updates while installing the app from TF.  Don't need to download the updates after installing the app. Updates should be downloaded only when the new updates are available and the user has already installed the app on his phone.

Here is my code

```
  useEffect(() => {
    checkForAppUpdate();
  }, []);

 const checkForAppUpdate = async () => {
    try {
      const updateInfo = await HotUpdater.checkForUpdate({
        source: SOURCE,
        requestHeaders: {
          // Add any necessary request headers here
        },
      });

      if (!updateInfo) {
        return {
          status: 'UP_TO_DATE',
        };
      }

      await HotUpdater.runUpdateProcess({
        source: SOURCE,
        requestHeaders: {
          // Add any necessary request headers here
        },
        reloadOnForceUpdate: true, // Automatically reload the app on force updates
      });

      return updateInfo;
    } catch (error) {
      console.error('Failed to check for update:', error);
      return null;
    }
  };

```

https://github.com/user-attachments/assets/7596b2a9-710b-46d2-9ebe-dda24246c606

### LINKED ISSUE #166: Feature Request: Add Support for Update Messages
#### Problem
No way to include a user-facing message (e.g., changelog) with updates in `hot-updater`, unlike CodePush’s `--description`.

#### Solution
1. Add `-m, --message <message>` to `npx hot-updater deploy` for update notes.
   - Example: `npx hot-updater deploy -p android -t 1.0.0 -m "Bug fixes"`
2. Add `message?: string` to `UpdateInfo` interface:
```typescript
   export interface UpdateInfo {
     id: string;
     shouldForceUpdate: boolean;
     fileUrl: string | null;
     fileHash: string | null;
     status: UpdateStatus;
     message?: string;
   }
```

#### Benefits
- Show users what’s in the update.
- Matches CodePush functionality.

#### Example
Deploy: `npx hot-updater deploy -p ios -t 1.2.0 -m "Fixed crashes"`
App: `alert(updateInfo.message)` → "Fixed crashes"

### LINKED ISSUE #167: WARN  `new NativeEventEmitter()` was called with a non-null argument without the required `addListener` method.
WARN  `new NativeEventEmitter()` was called with a non-null argument without the required `addListener` method.
 WARN  `new NativeEventEmitter()` was called with a non-null argument without the required `removeListeners` method.

how to remove this warning thanks ?

### LINKED ISSUE #176: Security related doubts (Supabase bucket & database are public)
Hi, firstly I wanted to thank you for this amazing solution which so simple and effective.
I'm new to react native and tried hot-updater with supabase provider and it got implemented in a single shot without any issues.

Now I've a few doubts related to security. For now the supabase bucket is public and if I make it private, will this implementation works or there are changes required in app for the same to let it work as expected.
Second, I just got a mail from Supabase security advisor to enable RLS in the database, If I do so, If there is anything additional needs to be done in the app implementation or it will work.

Please guide me through.

gronxb:
# HotUpdater Release Notes

## 🚀 New Features & Improvements**

1. **Storage Default Settings Updated**
* Supabase Storage now defaults to private.
* Cloudflare R2 now defaults to private as well.
* AWS was already using private by default.

2. **Signed URL-based Bundle Download**
* In Supabase, edge functions now download bundles using a signed URL.
* In Cloudflare, workers now download bundles using a signed URL (JWT-based).
* In AWS, Lambda@Edge now download bundles using a signed URL (JWT-based).

3. **Channel Support Added**
* If no channel is specified, the default is production.
* Example: Deploy to the dev environment using: `pnpm hot-updater deploy -c dev`

* To register a channel for an app, set releaseChannel in `hot-updater.config.ts` and build a new native app.
     * The initial bundle will be applied with the specified channel.
     * This ensures future updates are applied to the correct channel.

Configuration Example (`hot-updater.config.ts`)
```tsx
import { metro } from "@hot-updater/metro";
import { defineConfig } from "hot-updater";
import "dotenv/config";

export default defineConfig({
  releaseChannel: "dev", // here
  build: ...,
  storage: ...,
  database: ...,
});

```

4. **Updated Native Apps Won’t Reapply Old Updates**
* If a new native app build is created for the same target app version, it will not reapply previous updates.
* **Example:**
After deploying an update with HotUpdater, a newly uploaded TestFlight version (even with the same app version) will not download old bundles.

Closed #166
Closed #167 
Closed #86 
Closed #155
Maybe #150
Closed #176


packages/hot-updater/package.json
```diff
@@ -53,6 +53,7 @@
   },
   "dependencies": {
     "@clack/prompts": "^0.9.0",
+    "typescript": "5.8.2",
```
Embraser01, line 56: I didn't see the usage of this dependency, shouldn't it be in the devDependencies section?

Embraser01 (commented):
Hi 👋 
Thank you for this project and those new features!

Looking at the changes, it seems like the `channel` feature is made so that a specific app can't change the channel of the app during runtime. Would it be possible to have this possibility?

Our use case is to be able to switch environment directly from inside the app

wbercx:
> Our use case is to be able to switch environment directly from inside the app
I really enjoyed this about CodePush. I'd tell customers about a hidden feature in the app to switch to a staging channel, where they could test what we're building for them.

valeriobelli:
> Hi 👋 Thank you for this project and those new features!
> 
> Looking at the changes, it seems like the `channel` feature is made so that a specific app can't change the channel of the app during runtime. Would it be possible to have this possibility?
> 
> Our use case is to be able to switch environment directly from inside the app

Hello @Embraser01 👋🏽 

I see this requirement, but let's help @gronxb close the release without further non-strictly necessary development. March 31st is close, and this release can help many companies and individual developers move further with the switch.

What do you think?

gronxb:
@Embraser01 

Yes, for now, the priority is to ensure that this release is successfully merged.
Would you mind opening an issue for your suggestion?

Thanks for helping coordinate things, @valeriobelli !

valeriobelli:
> Totally make sense, this PR already contains a lot 🎉
> 
> Here is the related issue: #207

Thanks for your understanding, @Embraser01. Looking forward to seeing that feature shipped into Hot Updater. 🙌🏽

packages/hot-updater/package.json
gronxb: here https://github.com/gronxb/hot-updater/blob/08e98fcec7db107bd774acaa760fbc84d0f4c03e/packages/hot-updater/src/utils/transformTsEnv.ts

plugins/aws/iac/cloudfront.ts
```diff
@@ -0,0 +1,390 @@
...
+        },
+        LambdaFunctionAssociations: {
+          Quantity: 1,
+          Items: [
+            {
+              EventType: "origin-request",
+              LambdaFunctionARN: options.functionArn,
+            },
+          ],
+        },
```
valeriobelli, line 296: Hello @gronxb, is attaching the Lambda function to the default behavior necessary? As far as I can see from [here](https://github.com/gronxb/hot-updater/pull/168/files#diff-31d977323e2ff3e8c6aaddddb30328aab9cae5474aa822bd8edeed9bcd64ab2eR80-R82), the Lambda is passing through the request, but as far as I know, it is non strictly required.

Let's say that we have the bucket `foo-bar` in `us-east-1`, a CloudFront distribution listening to the record `https://hot-updater.foo-bar.com`, and we want to access the object key `production/x-y-z/bundle.zip`. The related S3 URL will be `https://foo-bar.s3.us-east-1.amazonaws.com/production/x-y-z/bundle.zip`, which can be converted to `https://hot-updater.foo-bar.com/production/x-y-z/bundle.zip` given we attached the S3 bucket as origin without any additional resources (e.g., Lambda) in the middle.

I might be missing something; therefore, please resolve this comment. 😊

gronxb: You’re right. I no longer need this after finishing the refactoring. 😅

gronxb: https://github.com/gronxb/hot-updater/pull/168/commits/9280c573882c246f26bff6046c1ebcb30d6dfeb0

PR #191, opened 2025-03-18
Adding Google Cloud Version

The-Gopher:


plugins/gcs/src/gcsDatabase.ts
gronxb, line 1: It looks like this was sourced from AWS. Are you planning to use Google Cloud Storage instead?

The spec has been updated recently — would it be possible to reimplement this based on the new spec?

Sorry for the trouble, and thank you so much in advance! 🥹

gronxb (changes requested):

plugins/gcs/src/gcsDatabase.ts
The-Gopher: Yes, this is very much a WIP and I very stole this from the s3 version.

I'm assuming I can just diff the version I branched from and main and see the changes.

The-Gopher:
Depends on the refactor from https://github.com/gronxb/hot-updater/pull/265

gronxb:
Hello, and thank you sincerely for all the work you’ve been doing. After careful consideration, I’d like to share some thoughts. I truly appreciate your efforts, and I believe that supporting Google Cloud in the future will be important as well.

However, I’d like to share a limitation I’ve encountered while managing an s3Database. Initially, it served well as a lightweight infrastructure solution. It was a valid choice until now. But as we try to track metrics like user activity and crash rates, the limitations have become clear.

Reading JSON files from S3 has never been an issue. However, when we try to record active sessions and crash reports via POST requests from our app, S3’s limitations with concurrent writes and cost efficiency become apparent. Because of this, we’re planning to add support for DynamoDB in the future.

As for GCP, I’m concerned that relying solely on blobDatabase may cause functional limitations down the line.

For the sake of future scalability, I believe that GCP should also provide support for a lightweight database solution, similar to how services like Cloudflare D1, Supabase Postgres, and AWS DynamoDB are currently supported.

I sincerely appreciate your support and hope you will kindly consider this suggestion.

The-Gopher:
Sorry for the late reply. This email got lost in the folder of random stuff
from GitHub I ignore.

I totally agree that the blob database is not ideal for a scalable version
it is just a method which works in our current environment without having
to start using a new service. I'm the only tech person at my company so I'm
very wary of onboarding a new technology to keep my cognitive load under
control.

Longer term I'm planning on investigating other database options but for
now I just want a basic version which can handle our relatively low volume
use case.

Great project and I'm hoping to get our integration in prod soon.

-Mike

gronxb:
How about trying out Firebase, which can be connected to GCP?

https://gronxb.github.io/hot-updater/guide/providers/4_firebase.html

gronxb:
We are closing it due to the discontinuation of the support plan. Firebase can serve as an alternative for now, and it may be replaced with a standalone option in the future. Thank you very much for all your efforts.

PR #200, opened 2025-03-22
Define workspace TypeScript version in settings

valeriobelli:
Setting the TypeScript version in the settings should allow for more consistency during development and make it less susceptible to the internal version of VS Code.


gronxb (approved):
Thanks !!

PR #201, opened 2025-03-22
feat(pocketbase): add pocketbaseStorage and pocketbaseDatabase plugin

jensneuber:
This adds a storage and a database plugin for PocketBase ( https://pocketbase.io ).

- a sample docker-compose file is included, also a matching schema for a `bundles` collection
- a S3 backend for Pocketbase is optional

ToDo: 
- tests for `pocketbaseStorage` are missing


plugins/pocketbase/example/pb_schema.json
```diff
@@ -0,0 +1,180 @@
...
+        "max": 0,
+        "min": 0,
+        "name": "fileUrl",
+        "pattern": "",
+        "presentable": false,
+        "primaryKey": false,
+        "required": false,
+        "system": false,
+        "type": "text"
+      },
```
gronxb, line 141: `fileUrl` is no longer in use.

A new field, `channel`, has been introduced.

Unfortunately, due to the spec changes, it looks like the previous implementation needs to be revised accordingly.

Feel free to reach out if you have any questions — I’d be happy to help further.

jensneuber: I am currently working on a pocketbase "update check" endpoint to return the required fileUrl.
Should be ready soon :)

gronxb:
I’ve given this a lot of thought.

Since I’m managing the open source project on my own, it’s hard for me to maintain everything when too many things are added.

So, going forward, I plan to support only well-known vendors as providers.

Currently, I’m supporting Supabase, Firebase, AWS, and Cloudflare, but in the future, I probably won’t include anything unless it’s something like Azure or GCP.

If something is really necessary, you’ll be able to publish a plugin yourself, and I’ll treat it as a community project and mention it in the documentation.

Sorry and Thank you

PR #215, opened 2025-03-27
add server plugin for postgresql

miso-kyoungminkim:
Due to the deprecation of CodePush, I plan to use hot-updater as an alternative.
Hot-updater provides 3 types of workers : lambda, cloudflare, supabase.
However, I require a worker that runs on nodejs runtime and uses postgresql.
For this reason, I have created this PR.

I would appreciate your review and any feedback you may have. Thank you!

PR #225, opened 2025-03-28
feat(standalone): basic node-server (postgres)

gronxb:
Related #215


maelp:
This seems quite a bit easier to setup, is this working now?

maelp:
It could even be nice with a Drizzle+Sqlite backend, so we don't even need to rely on Postgres

roshangm1:
@gronxb The CLI should have the option to use this too!

gronxb:
Thanks everyone for your input. I think it will probably be rewritten. I’ve already abstracted everything in my head, but I just don’t have enough time. I’m planning to scaffold the project simply through the CLI.

PR #242, opened 2025-04-01
fix(android): keep a stable minBundleId on Android

flodlc:
* As-is
getMinBundleId is based on the APK’s modification date, so when the app is freshly installed from the Play Store, it always gets updated to the latest version. As a result, update bundles created before the installation are ignored.
* To-be
getMinBundleId is fixed based on the build time, ensuring consistency regardless of the installation date.


gronxb (approved):
thanks !!!!!!🙏🙏

PR #250, opened 2025-04-03
Allow to specify a custom output path for bundle.zip

valeriobelli:
This Pull Request introduces the possibility of specifying the path where to store the `bundle.zip` before uploading it. After this addition, Hot Update will allow concurrently bundling of Android and iOS platforms.

### DIFF
--- a/packages/hot-updater/src/commands/deploy.ts
+++ b/packages/hot-updater/src/commands/deploy.ts
@@ -90,18 +90,7 @@ export const deploy = async (options: DeployOptions) => {
         })
       : null);
 
-  const outputPath =
-    options.bundleOutputPath ??
-    (options.interactive
-      ? (
-          await p.text({
-            defaultValue: cwd,
-            initialValue: cwd,
-            message: "Bundle.zip output path (e.g., ./dist/bundle/ios)",
-            placeholder: cwd,
-          })
-        ).toString()
-      : cwd);
+  const outputPath = options.bundleOutputPath ?? cwd;
 
   if (p.isCancel(targetAppVersion)) {
     return;
--- a/packages/hot-updater/src/commands/deploy.ts
+++ b/packages/hot-updater/src/commands/deploy.ts
@@ -1,4 +1,4 @@
-import fs from "node:fs";
+import fs from "fs";
 import semverValid from "semver/ranges/valid";
 
 import open from "open";
--- a/packages/hot-updater/src/commands/deploy.ts
+++ b/packages/hot-updater/src/commands/deploy.ts
@@ -1,4 +1,4 @@
-import fs from "fs";
+import fs from "node:fs";
 import semverValid from "semver/ranges/valid";
 
 import open from "open";
@@ -26,12 +26,13 @@ import { getBundleZipTargets } from "@/utils/getBundleZipTargets";
 import { printBanner } from "@/utils/printBanner";
 
 export interface DeployOptions {
-  targetAppVersion?: string;
-  platform?: Platform;
+  bundleOutputPath?: string;
+  channel: string;
   forceUpdate: boolean;
   interactive: boolean;
-  channel: string;
   message?: string;
+  platform?: Platform;
+  targetAppVersion?: string;
 }
 
 export const deploy = async (options: DeployOptions) => {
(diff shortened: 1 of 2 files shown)


packages/hot-updater/src/commands/deploy.ts
```diff
@@ -260,10 +280,10 @@ export const deploy = async (options: DeployOptions) => {
     }
     p.outro("🚀 Deployment Successful");
   } catch (e) {
-    await databasePlugin.onUnmount?.();
```
gronxb, line 263: It seems necessary because `finally` is not reached due to `process.exit`.

packages/hot-updater/src/commands/deploy.ts
```diff
@@ -1,4 +1,4 @@
-import fs from "fs";
+import fs from "node:fs";
```
gronxb, line 1:
```suggestion
import fs from "fs";
```

packages/hot-updater/src/commands/deploy.ts
```diff
@@ -89,6 +90,19 @@ export const deploy = async (options: DeployOptions) => {
...
+    (options.interactive
+      ? (
+          await p.text({
+            defaultValue: cwd,
+            initialValue: cwd,
+            message: "Bundle.zip output path (e.g., ./dist/bundle/ios)",
+            placeholder: cwd,
+          })
+        ).toString()
+      : cwd);
```
gronxb, line 104:
```suggestion
  const outputPath =
    options.bundleOutputPath ?? cwd;
```

It doesn’t seem important enough to require keyboard input, so I’ll exclude it from the interaction mode.

gronxb (changes requested):

packages/hot-updater/src/commands/deploy.ts
valeriobelli: Hey @gronxb, I can change it without any issue. 😁 
However, out of curiosity, wouldn't be safer to declare `node:` in the imports? This will ensure to resolve node modules instead of an eventual modules contained in the `node_modules`.

packages/hot-updater/src/commands/deploy.ts
valeriobelli: I agree. I'll drop this change 👍

packages/hot-updater/src/commands/deploy.ts
valeriobelli: Good catch; I completely overlooked `process.exit(1)`. 😵

packages/hot-updater/src/index.ts
```diff
@@ -46,6 +46,12 @@ program
   .addOption(
     new Option("-f, --force-update", "force update the app").default(false),
   )
+  .addOption(
+    new Option(
+      "-bo, --bundle-output-path <bundleOutputPath>",
+      "the path where the bundle.zip will be generated",
+    ),
+  )
```
gronxb, line 54: Personally, I find “-bo” a bit unattractive—how about “-o” instead?

gronxb (changes requested):

packages/hot-updater/src/commands/deploy.ts
gronxb: I agree with your point, but I was thinking about the distant future, where it should also work with runtimes like Deno and Bun. Of course, using the node: protocol works in both Deno and Bun, but since it’s handled through a compatibility layer, I wanted to avoid a hard dependency on Node—even though we’re currently targeting Node.

packages/hot-updater/src/commands/deploy.ts
```diff
@@ -260,10 +269,11 @@ export const deploy = async (options: DeployOptions) => {
     }
     p.outro("🚀 Deployment Successful");
   } catch (e) {
-    await databasePlugin.onUnmount?.();
+    await fs.promises.rm(bundlePath, { force: true });
```
gronxb, line 272:
```suggestion
    await databasePlugin.onUnmount?.();
    await fs.promises.rm(bundlePath, { force: true });
```

gronxb (approved):

gronxb:
Thanks !!

PR #265, opened 2025-04-08
chore (aws): split s3Database into createBlobDatabase and s3Database

The-Gopher:
Moves the logic out for maintaining a database on a blob type service to a separate file. The blob database only needs functions to list, upload, download, delete be specific to the data source.


The-Gopher:
I'm not a AWS expert but is the handling of cloud-front invalidation required? 

The only thing reading this data is a lambda function do they use the cache or do they access the s3 data directly?

gronxb:
S3 and CloudFront are tightly integrated: when a file is uploaded to S3, it is globally redistributed through CloudFront and cached. As a result, we configure S3 to retain its cache for up to one year and rely entirely on CloudFront’s caching behavior. To reduce costs, we avoid direct access to S3 and only serve content through CloudFront.

The reason we needed the cloudfrontDistributionId was to selectively invalidate the cache only for changed files, since the cache is set to last for a year.

This strategy helps minimize costs by leveraging long-term caching unless a file is updated.

I’m not an expert in GCP either, so I’m wondering—what is the equivalent setup in GCP to achieve the same result?

The-Gopher:
I was confused by the non-dictionary logic - I now see the link returned to the device to download the bundle is from cloud front, not s3 directly.

GCP does, we don't currently use it so I'm not 100% sure how it works but it is probably similar to CloudFront. My guess is if/when GCP Cloud CDN is added we can use the same hook mark the paths as invalid.

gronxb:
I believe this PR is valid, so I’ll review and merge it soon.

plugins/aws/src/s3Database.ts
```diff
@@ -231,264 +146,43 @@ export const s3Database = (
     region: s3Config.region,
   });
 
-  // Map for O(1) lookup of bundles.
-  const bundlesMap = new Map<string, BundleWithUpdateJsonKey>();
-  // Temporary store for newly added or modified bundles.
-  const pendingBundlesMap = new Map<string, BundleWithUpdateJsonKey>();
-
-  const PLATFORMS = ["ios", "android"] as const;
+  const litemItems = (prefix: string) =>
```
gronxb, line 149:
```suggestion
  const listItems = (prefix: string) =>
```

plugins/plugin-core/src/createBlobDatabasePlugin.ts
```diff
@@ -0,0 +1,380 @@
...
+ */
+export const createBlobDatabasePlugin = (
+  name: string,
+  listObjects: ListObjectsFn,
+  loadObject: LoadObjectFn,
+  uploadObject: UploadObjectFn,
+  deleteObject: DeleteObjectFn,
+  invalidatePaths: InvalidatePathsFn,
+  hooks?: DatabasePluginHooks,
+) => {
```
gronxb, line 61: I think we should pass them as an object since there are too many arguments.

`createBlobDatabasePlugin({ name, listObjectsFn: () ~~})`

gronxb (changes requested):

plugins/plugin-core/src/createBlobDatabasePlugin.ts
The-Gopher: Fixed

plugins/aws/src/s3Database.ts
The-Gopher: Fixed

gronxb (approved):

Issue #278, opened 2025-04-11
reload() on Android cause crash


OskarAtJoint:
Same as https://github.com/gronxb/hot-updater/issues/150 , but cant reopen 

## Environment
- React Native Version: 0.73.9 
- `hot-updater` Version: 0.16.0
- Platform: android

gronxb:
I want to solve this issue, but I can’t reproduce it. It would be helpful if someone could provide a repro that consistently reproduces the problem.

When providing a repro, I think it’s fine to exclude the environment setup. Since the issue is related to reload, it should be enough to just test the reload behavior.

OskarAtJoint:
think this fixes it, loading Asynchronous
```diff
diff --git a/node_modules/@hot-updater/react-native/android/src/main/java/com/hotupdater/ReactIntegrationManagerBase.kt b/node_modules/@hot-updater/react-native/android/src/main/java/com/hotupdater/ReactIntegrationManagerBase.kt
index ccfc721..bf2581a 100644
--- a/node_modules/@hot-updater/react-native/android/src/main/java/com/hotupdater/ReactIntegrationManagerBase.kt
+++ b/node_modules/@hot-updater/react-native/android/src/main/java/com/hotupdater/ReactIntegrationManagerBase.kt
@@ -16,7 +16,7 @@ open class ReactIntegrationManagerBase(
                 JSBundleLoader.createAssetLoader(
                     context,
                     bundleFileUrl,
-                    true,
+                    false,
                 )
         } else {
             bundleLoader = JSBundleLoader.createFileLoader(bundleFileUrl)
```
ref:
https://github.com/microsoft/react-native-code-push/blob/6015386f22c61417cd2b2ae2860a9ad69c37f643/android/app/src/main/java/com/microsoft/codepush/react/CodePushNativeModule.java#L118

gronxb:
Could you create a pull request for this?

Thank you!

PR #281, opened 2025-04-14
loading assetloader synchronously, cause crash, load async instead

OskarAtJoint:
https://github.com/gronxb/hot-updater/issues/278


gronxb (approved):
your best. Thanks!

imki123:
Is it published?
I’ve updated to the latest version, but the changes don’t seem to be applied.

In my case, the third parameter in `JSBundleLoader.createAssetLoader(...)` is set to `true`.

PR #297, opened 2025-04-24
feat: added maxRetries config

Hariomagr:
**Description**
This package reloads the React Native JS bundle on forceUpdate. However, we've encountered a potential issue that can cause infinite reload loops if the backend update check API (updatedInfo) always returns a "new update available" response, even if the app is already up to date.

This typically happens if the API is incorrectly configured or the backend has an issue.

**The Problem**
- The initHotUpdater() function runs in useLayoutEffect.
- It reloads the JS bundle when forceUpdate === true.
- There’s no limit on how many times this retry can happen.
- So, if the backend always returns "update available", the app will infinitely reload the bundle after every attempt.

**Proposed Solution**
Introduce a maxRetry configuration in CheckForUpdateConfig, e.g.:
`interface CheckForUpdateConfig {
  maxRetries?: number; // Optional retry limit
}`
We track the retry attempts persistently on the native side (both Android and iOS), so the count survives bundle reloads.


packages/react-native/android/src/main/java/com/hotupdater/HotUpdater.kt
```diff
@@ -26,6 +26,16 @@ class HotUpdater : ReactPackage {
         listOf(HotUpdaterModule(context)).toMutableList()
 
     companion object {
+        private var count: Int = 0
+
```
regalstreak, line 30: Could it be named retryCount or something more semantic?

packages/react-native/ios/HotUpdater/HotUpdater.mm
```diff
@@ -161,7 +169,7 @@ - (void)cleanupOldBundlesAtDirectory:(NSString *)bundleStoreDir {
         NSLog(@"Failed to list bundle store directory: %@", error);
```
regalstreak, line 169: can remove new line changes

packages/react-native/src/native.ts
```diff
@@ -115,3 +115,17 @@ export const getChannel = (): string | null => {
   const constants = HotUpdaterNative.getConstants();
   return constants?.CHANNEL ?? HotUpdater.CHANNEL ?? null;
 };
+
+/**
+ * Fetches the execution count
+ */
+export const getExecutionCount = (): Promise<number> => {
+  return HotUpdaterNative.getExecutionCount();
```
regalstreak, line 123: do we add a trycatch here?

packages/react-native/src/native.ts
```diff
@@ -115,3 +115,17 @@ export const getChannel = (): string | null => {
...
+};
+
+/**
+ * Increments the execution count
+ */
+export const updateExecutionCount = () => {
+  HotUpdaterNative.updateExecutionCount();
```
regalstreak, line 130: here too, maybe name this increment?

packages/react-native/src/wrap.tsx
```diff
@@ -143,7 +150,15 @@ export function wrap<P extends React.JSX.IntrinsicAttributes = object>(
       }, [progress]);
 
       useLayoutEffect(() => {
-        initHotUpdater();
+        (async () => {
+          if (reloadOnForceUpdate && maxRetries) {
+            const count = await getExecutionCount();
```
regalstreak, line 155: rename to retryCount

packages/react-native/android/src/oldarch/HotUpdaterModule.kt
```diff
@@ -85,6 +85,16 @@ class HotUpdaterModule internal constructor(
         return constants
     }
 
+    @ReactMethod
+    override fun getExecutionCount(promise: Promise) {
+        promise.resolve(HotUpdater.getCount())
+    }
+
+    @ReactMethod
+    override fun updateExecutionCount() {
```
regalstreak, line 94: rename to retry over execution?

regalstreak (approved):

packages/react-native/android/src/main/java/com/hotupdater/HotUpdater.kt
Hariomagr: can be executionCount?

packages/react-native/src/native.ts
Hariomagr: would it make sense here? since it's only incrementing and getting the data?

packages/react-native/src/wrap.tsx
Hariomagr: executionCount would be better imo, we are using executionCount to check for maxRetries but is it not retry count

gronxb:
I’d like to discuss this with you all.

Instead of increasing the count in JS (since crash rates increase as JS-native communication becomes more frequent),

how about passing `maxRetries` to `updateBundle`, and on the native side, mapping each URL (excluding query strings) to `updateCount` in a one-to-one way, storing it using `HotUpdaterPrefs` so that it can be persistently saved per app version?

Then, if the retry count for a given URL exceeds `maxRetries` in `updateBundle`, how about having it return `Promise.reject()`?

gronxb:
I realized that there’s no need to store data per app version in HotUpdaterPrefs; it only needs to exist at runtime.
However, I think it’s valid that zipUrl and retryCount are linked.

Hariomagr:
Thanks for the feedback! Just to clarify the purpose behind the maxRetries config:

This is primarily to prevent infinite reload loops that can occur after updateBundle(), when a new bundle is fetched from the backend and applied immediately.

For downloaded bundles, execution logic is already handled on the native side during bridge initialization

Although I do agree with your point about updateExecutionCount being trackable on the native side during updateBundle.

gronxb:
I understand what you're trying to achieve.

However, the `reload()` is only triggered when `checkForUpdate` returns `shouldForceUpdate: true` **and** `updateBundle` completes successfully.  
In other words, if `updateBundle` fails, `reload()` won't be triggered even if `shouldForceUpdate` is `true`.

What I'm concerned about is that exposing a method like `getExecutionCount()` and documenting it could introduce a learning curve for developers.  
While it's used internally via the wrapper, developers implementing a custom update flow would still need to understand this context.

If we handle everything inside `updateBundle` as you suggested, the infinite loop prevention works out of the box, and developers don’t need to worry about anything.  
There would be no need to modify the JS code either.

Hariomagr:
> If we handle everything inside updateBundle as you suggested, the infinite loop prevention works out of the box, and developers don’t need to worry about anything.
There would be no need to modify the JS code either.

I do agree with this too. How do we want to send the retry count to the native side though? I was thinking to have the init function which set these params to the native side (maxRetry and other configs we want to send), but then again it has to be before anything is called (before the initHotUpdater). Or other way is sending the param to the updateBundle, then again this will not be scalable.

gronxb:
I think it's okay to send `maxRetries` to `updateBundle`. Why do you think it hurts scalability?

Even if not now, I plan to send an object later on.
updateBundle({
    maxRetries: 5,
    bundleId: "",
    downloadUrl: "",
});

On the JS side, I thought it would be fine to change the type signature of `updateBundle` and pass `maxRetries` from the `HotUpdaer.wrap` to `updateBundle`.

Hariomagr:
> I think it's okay to send maxRetries to updateBundle. Why do you think it hurts scalability?

in future, there might be a need to pass some other param to native side. I cannot think of it anything now, but there is always a possibility. Anyways for now, i'll just make these changes

Hariomagr:
@gronxb i have added and tested the changes.

gronxb:
Could you run `ktlint android --format` in the `packages/react-native` directory?

Hariomagr:
hey @gronxb, can we merge this if everything looks good?

gronxb (approved):
Thanks for your contribution! From what I can tell, it looks good. However, since this is a version that requires a review, we should bump the minor version. There’s a pending package release, so after that goes out, I’ll include this with version 0.17.0.

Did the e2e tests go well?

Hariomagr:
thanks, yes, all the e2e tests passed.

gronxb:
Although it hasn’t been released yet, the iOS version is not working, so the release has been delayed🤔

Hariomagr:
let me check this quickly. Last time i tried, it was working

gronxb:
I’ll look into this too — it seems like the crash is happening in updateBundle.

PR #300, opened 2025-04-25
feat(rnef): supports `@hot-updater/repack` / `@hot-updater/rnef`  build plugin

gronxb:
스크린샷 2025-04-26 오전 3 20 08

And Support Repack


maxinminax:
It will support repack without rnef?

gronxb:
Yes. I have currently set the `@hot-updater/metro`. plugin to `@hot-uodater/bare` for this, and you can use a combination of the bare and repack plugins.

PR #311, opened 2025-04-30
docs: update sentry

Arthur-Noh:
Hello,
I noticed that something was missing from the document, so I made some corrections.
I would appreciate it if you could take a look.


gronxb (approved):
Thanks !!

PR #315, opened 2025-05-06
chore: Remove wrongly injected banner in cjs bundles

valeriobelli:
A banner with `import` syntax is injected in CommonJS modules, breaking the evaluation. The following are some evidences

image

image


gronxb (approved):
ah..! i missed it. Thanks 🥰

PR #365, opened 2025-06-01
Stored bundle path should be separated by channel

### LINKED ISSUE #364: [android]  Bundle paths should be separated by channel
I have a case where one `versionName` has two hot-updater channels

When trying to update from one to the other (I'm using Firebase App Tester), the bundle from the previous build is retained. This breaks the update source url as the value from `HotUpdaterConstants.OVER_THE_AIR_CHANNEL` does not match `HotUpdaterNative.getConstants().CHANNEL` (and the first one is prioritized over the latter)

Even if both are the same, there might still be an update issue as it would provide a non-matching bundle ID

This is a case that I've carried over from using microsoft's code push, as it doesn't pollute the list of app builds that much (for iOS, it'd be multiple version numbers, but the same Marketing Version string)

Is this an intentional behavior, or would it be possible to separate them?
I can try to make a pull request to solve this issue

I've only tested this so far with Android but I assume it'd be the same on ios

Thank you!

EDIT:

It seems like this can be easily resolved by attaching a channel name to the key `HotUpdaterBundleURL` instead of using as is

EDIT2:

```
"@hot-updater/react-native": "^0.18.1",
    "react-native": "^0.74.3",
```

marqroldan:
Closes #364

This PR should allow multiple release channels with the same version name / marketing version


gronxb:
Thanks for your contribution!

I think it might be a good idea to consider modifying the constructor in `VersionedPreferencesService`.

The following four code snippets might help you understand the context better:

https://github.com/gronxb/hot-updater/blob/3e469b6eb7e0f4069b2dda5bbc7b3294384d2308/packages/react-native/android/src/main/java/com/hotupdater/VersionedPreferencesService.kt#L34-L40  
https://github.com/gronxb/hot-updater/blob/3e469b6eb7e0f4069b2dda5bbc7b3294384d2308/packages/react-native/android/src/main/java/com/hotupdater/HotUpdaterFactory.kt#L33  
https://github.com/gronxb/hot-updater/blob/3e469b6eb7e0f4069b2dda5bbc7b3294384d2308/packages/react-native/ios/HotUpdater/Internal/VersionedPreferencesService.swift#L26-L29  
https://github.com/gronxb/hot-updater/blob/3e469b6eb7e0f4069b2dda5bbc7b3294384d2308/packages/react-native/ios/HotUpdater/Internal/HotUpdaterImpl.swift#L40-L42

gronxb:
cd packages/react-native

could you run `ktlint android --format` ?

gronxb:
If there’s any test evidence, we might be able to merge it quickly! Thanks a lot!

marqroldan:
I'll provide a test evidence within the hour, I got it to work on android
But I'm having an issue on ios-- it downloads the build but encounters an error here: `bundleStorage.updateBundle(bundleId: bundleId, fileUrl: fileUrl)` (result is .failure)

Will check first if this also occurs on the original package (it was working fine on 0.17 i believe)

gronxb:
The codebase has likely changed a lot due to the transition to Swift between versions 0.17.0 and 0.18.0.
However, it was working fine when I tested it. That’s strange… hmm.

gronxb:
Thank you, I removed the Swift code you mentioned.

#372

PR #384, opened 2025-06-05
`app-version` shows naive native app version

mym0404:
# Overview

Users would expect `app-version` command shows the current version of inner native platforms. However, it has generated unexpected **default target version** even `null` from iOS.

The main cause of the behavior returning `null` from iOS was `find-up-simple` function has been failed to search `Info.plist` with the wrong current directory. It searches `Info.plist` bottom-up not top-down.

I have fixed this logic with returning `getNativeAppVersion` instead along with extra refactors for the utils associated with version logics.

# Changes

- [x] Switch `app-version` command logic into returning real current native app version.
- [x] Fix iOS app version parsing logic with `Info.plist`
- [x] Remove `find-up-simple` dependency
- [x] Refactor version utils placed in `hot-updater/src/utils/version`
- [x] Fix broken tests

---

Any suggestion welcome

### DIFF
--- a/packages/hot-updater/src/utils/version/getAndroidVersion.ts
+++ b/packages/hot-updater/src/utils/version/getAndroidVersion.ts
@@ -31,10 +31,10 @@ export const getAndroidVersion = async ({
   parser: AndroidVersionParser | AndroidVersionParser[];
   validateWithSemver?: boolean;
 }): Promise<string | null> => {
-  const strategies = Array.isArray(parser) ? parser : [parser];
+  const parsers = Array.isArray(parser) ? parser : [parser];
 
-  for (const strategy of strategies) {
-    const parsedVersion = await AndroidVersionParsers[strategy]();
+  for (const parserKey of parsers) {
+    const parsedVersion = await AndroidVersionParsers[parserKey]();
 
     if (!parsedVersion) continue;
     if (validateWithSemver && !semverValid(parsedVersion)) continue;
--- a/packages/hot-updater/src/utils/version/getIOSVersion.ts
+++ b/packages/hot-updater/src/utils/version/getIOSVersion.ts
@@ -83,10 +83,10 @@ export const getIOSVersion = async ({
   parser: IOSVersionParser | IOSVersionParser[];
   validateWithSemver?: boolean;
 }): Promise<string | null> => {
-  const strategies = Array.isArray(parser) ? parser : [parser];
+  const parsers = Array.isArray(parser) ? parser : [parser];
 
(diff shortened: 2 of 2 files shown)


mym0404:
oh, I'll upload changeset

packages/hot-updater/src/utils/getIosAppTargetDirectoryName.ts
```diff
@@ -0,0 +1,15 @@
...
+export const getIosAppTargetDirectoryName = () => {
+  const iosDirectory = path.join(getCwd(), "ios");
+
+  const [xcodeprojPath] = globbySync("*.xcodeproj/project.pbxproj", {
+    cwd: iosDirectory,
+    absolute: false,
+    onlyFiles: true,
+  });
+
+  return xcodeprojPath?.split(".")?.[0] ?? null;
```
mym0404, line 14: should we have enhanced validation logic?

mym0404:
> I’m curious why you chose that strategy. Do you think other strategies will be added?

I kept `app.json` in Expo parser in mind. But I'm unsure if parsing version from the expo config is right.

packages/hot-updater/src/utils/getIosAppTargetDirectoryName.ts
gronxb: I think null would be enough👍

gronxb (approved):
Oh, you’re right. In the Expo CNG environment, I should check the app.json file.

mym0404:
yup, also `app.config.[t|mj|cj]s`

packages/hot-updater/src/utils/version/getAndroidVersion.ts
```diff
@@ -0,0 +1,46 @@
...
+    const versionNameMatch = buildGradleContent.match(
+      /versionName\s+['"]([^"]+)['"]/,
+    );
+    return versionNameMatch?.[1] ?? null;
+  } catch (error) {
+    return null;
+  }
+};
+
+const Strategy = {
```
gronxb:
```suggestion
const AndroidVersionStrategy = {
```
?

gronxb (changes requested):

packages/hot-updater/src/utils/version/getAndroidVersion.ts
mym0404: I see, the naming of `Strategy` feels exaggeration. I have changed!

gronxb (approved):
👍👍

PR #387, opened 2025-06-06
chore: restore expo example to "deployable" state

mym0404:
# Overview

The current `expo-52` example project is not working to test package commands usefully.

I don't have expertise in `.pnpm` like link system, but have made it just work on my machine at least and altered format of project into more like basic `create-expo-app` sample output.

The root cause of disorder seems to the wrong version selection algorithm to discover valid `metro` related packages in pnpm workspace. The invalid version of packages has been picked up from other old rn version example maybe?

- drop `rnx` dependencies
- fix `metro`, `metro-resolver` version in example project

# Test plan

- [x] I succeeded supabase deploy with `deploy` command


examples/expo-52/package.json
```diff
@@ -18,19 +18,14 @@
...
-    "@expo/metro-config": "^0.19.12",
+    "@babel/runtime": "^7.25.2",
     "@hot-updater/expo": "workspace:*",
     "@hot-updater/supabase": "workspace:*",
-    "@react-native/metro-config": "0.76.9",
-    "@rnx-kit/metro-config": "^2.0.1",
-    "@rnx-kit/metro-resolver-symlinks": "^0.2.1",
     "@types/react": "~18.3.12",
-    "babel-preset-expo": "12.0.11",
-    "dotenv": "^16.4.5",
```
gronxb, line 29: I think it might be necessary to inject values into the hot-updater config.

mym0404: yep, I just rolled back.

PR #394, opened 2025-06-07
Android Native Build

mym0404:


mym0404:
@gronxb I'm considering 2 options like the below. Which one do you prefer? Option 1 seems more clear but it'll break backward compatibility.

```ts
export default defineConfig({
  nativeBuild: expo.nativeBuild({
    android: {...},
    ios: {...},  
  }),
  build: expo.build({ enableHermes: true }),
  storage: ...
  database: ...
});
```

vs

```ts
export default defineConfig({
  nativeBuild: expoNativeBuild({
    android: {...},
    ios: {...},  
  }),
  build: expo({ enableHermes: true }),
  storage: ...
  database: ...
});
```

gronxb:
How about reusing the build plugin?
For example, we can set it up as `build: expo()`, but I think this expo plugin can also define native build specifications inside it.

mym0404:
> How about reusing the build plugin? For example, we can set it up as `build: expo()`, but I think this expo plugin can also define native build specifications inside it.

It's an option. Indeed, that also requires changing argument type of `expo` function itself.

gronxb:
The idea is to add the nativeBuild specification [here](https://github.com/gronxb/hot-updater/blob/708089df1304725c224f2bf234b53c9074fb35c4/plugins/plugin-core/src/types/index.ts#L46) and let each plugin define its own behavior based on it.

mym0404:
Yes we can add `nativeBuild` like function into `BuildPlugin`, I'm concerning about `expo` function usage have to be updated.

As far, the first argument of `expo(...)` is options for `build(bundle)`

mym0404:
@gronxb 

⸻

Although many features haven’t been added yet, I’ve confirmed that the Android APK and AAB builds are working in examples/v0.77.

I’d like to merge this much in advance to prevent divergence issues between different core types.

You can test it in the example folder using:

```
NODE_ENV=development npx hot-updater build:native -p android
```

If you have some time, please give it a try and review it.

packages/hot-updater/src/commands/build:native.ts
```diff
@@ -0,0 +1,297 @@
...
+    });
+
+    target.fingerprintHash = fingerprint.hash;
+    s.stop(`Fingerprint(${platform}): ${fingerprint.hash}`);
+
+    if (!target.fingerprintHash) {
+      p.log.error(`Failed to calculate fingerprint of ${platform}`);
+      return;
+    }
+  }
```
gronxb: How about using `createFingerprintJson` to update fingerprint.json and start the process? But only when the config.updateStrategy is set to "fingerprint".

https://github.com/gronxb/hot-updater/blob/40d28c2a7dfbf787a3987fa43b946ca275f63411/packages/hot-updater/src/utils/fingerprint/index.ts#L86

mym0404: are there any other scripts to run before native build? syncing fingerprint and expo prebuild are sufficient?

gronxb: Yes, that’s right. The fingerprint is common, and the prebuild is probably only for Expo.

packages/hot-updater/src/commands/build:native.ts
```diff
@@ -0,0 +1,297 @@
...
+          taskRef.buildResult = await buildPlugin.nativeBuild({
+            platform: platform,
+            // inject native build function into plugins
+            // then plugin will run it with pre/post required steps for each framework
+            buildNativeArtifact: async () => {
+              await runNativeBuild({
+                platform,
+                config: config.nativeBuild,
+              });
+            },
```
gronxb: I think the flow of `buildNativeArtifact` is a bit confusing.

Currently, the flow assumes that the CLI injects `buildNativeArtifact` into the plugin and it's executed from there, but then the CLI calls the plugin again.

Since we decided to centralize the actual build logic, how about letting the build plugin define only the *before* and *after* behaviors of `nativeBuild`, and handle the rest in the CLI?

For example, we can define the plugin type like this:
```tsx
nativeBuild?: {
  pre?: () => Promise<void>;
  post?: () => Promise<void>;
}
```
Then the CLI can handle it like this:
```tsx
await buildPlugin?.nativeBuild?.pre?.();
await runNativeBuild({
  platform,
  config: config.nativeBuild,
});
await buildPlugin?.nativeBuild?.post?.();
```

packages/hot-updater/src/index.ts
```diff
@@ -133,4 +132,33 @@ program
     log.info(`iOS version: ${iosVersion}`);
   });
 
+// developing command gorups
+if (process.env["NODE_ENV"] === "development") {
```
gronxb, line 136: I was thinking about how to merge without exposing it to users, and this is just perfect. ⭐️

packages/hot-updater/src/commands/build:native.ts
mym0404: That's such cool. It doesn't break dependency inversion and looks rather clear. I'll apply it.

mym0404:
by the way, usage of catalog: pnpm versioning is safe for this package's publish settings? If publishing is not managed by pnpm itself, then catalog: won't translated to actual workspace package versions. @gronxb

gronxb:
You can try running `pnpm pack` on the one that depends on catalog and check the package.json inside. That way, you’ll know for sure. I’ve been meaning to look into catalog anyway, so this is great.

mym0404:
I've checked result of `pnpm pack` and it was translated into real version correctly.

packages/hot-updater/src/utils/fingerprint/index.ts
```diff
@@ -51,7 +94,7 @@ const ensureFingerprintConfig = async () => {
   const config = await loadConfig(null);
   if (config.updateStrategy === "appVersion") {
     p.log.error(
-      "The updateStrategy in hot-updater.config.ts is set to 'appVersion'. This command only works with 'fingerprint' strategy.",
+      "The updateStrategy in hot-updater.config.ts is set to 'appVersionu'. This command only works with 'fingerprint' strategy.",
```
gronxb:
```suggestion
      "The updateStrategy in hot-updater.config.ts is set to 'appVersion'. This command only works with 'fingerprint' strategy.",
```

plugins/bare/src/bare.ts
```diff
@@ -143,6 +143,7 @@ export const bare =
       enableHermes,
     } = config;
     return {
+      nativeBuild: {},
```
gronxb:
```suggestion
```

I think it’s okay to make this nullable as well.

packages/hot-updater/src/commands/build:native.ts
```diff
@@ -0,0 +1,303 @@
...
+
+    if (!target.appVersion) {
+      p.log.error(`Failed to retrieve native app version of ${platform}`);
+      return;
+    }
+  }
+
+  // const nativeBuildConfig = config.nativeBuild;
+
+  const outputPath = options.outputPath ?? path.join(cwd, "nativebuild"); // TODO any suggestion?
```
gronxb, line 111: How about using the default path that is generated by Android (per platform)?

gronxb (changes requested):

packages/hot-updater/src/commands/build:native.ts
mym0404: or it would be cool if we configure .hot-updater/output as a common result path for bundle or build for who wants to know or investigate what happens by their commands.

gronxb: ` .hot-updater/output` 

it's okay.

mym0404: Can I separate this work as a other PR? @gronxb

gronxb: yes

Issue #341, opened 2025-05-21
Sentry does not support artifact bundle


Kinqdos:
- [x] Infrastructure may have changed. Did you retry `hot-updater init`?
- [x] Is your Node version v20 or higher?
- [x] Are you aware that the update will only work properly if you `hot-updater deploy` after building the native app?

## Problem
I use the sentry plugin to upload sourcemaps, but I always get this error:

![Image](https://github.com/user-attachments/assets/1183dec9-790b-4fde-9c64-3c33b5e047ac)

## Steps to Reproduce
Use sentry plugin. This is my config code:
```typescript
withSentry(
        bare({entryFile: "index.tsx", enableHermes: true, sourcemap: true}),
        {
            url: process.env.SENTRY_URL,
            org: process.env.SENTRY_URL,
            project: process.env.SENTRY_PROJECT,
            authToken: process.env.SENTRY_AUTH_TOKEN
        }
    )
```

## Environment
- React Native Version: 0.79.2
- `hot-updater` Version: 17.0
- Platform: both
- OS: macos

gronxb:
What is the version of the Sentry CLI? I'm also curious about the versions of all Sentry-related packages.

Kinqdos:
It uses the one from its own dev dependency, which is 2.45.0. This is also the latest version atm.

gronxb:
Can you share your `metro.config.js` file?

gronxb:
Is the entry file really index.tsx? I’m just curious because in most projects it’s usually index.js.

Kinqdos:
Sure here is my metro.config.js
```javascript
const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');
const {withSentryConfig} = require("@sentry/react-native/metro");

/**
 * Metro configuration
 * https://reactnative.dev/docs/metro
 *
 * @type {import('@react-native/metro-config').MetroConfig}
 */
const config = {};

module.exports = withSentryConfig(mergeConfig(getDefaultConfig(__dirname), config));
```

`index.tsx` is correct, but I can share this too.

```tsx
import 'react-native-url-polyfill/auto';
import 'dayjs/locale/de' //Damit day.js auf deutsch ist
import "reflect-metadata";
import './src/extension/MMKV.ts'
import * as Sentry from "@sentry/react-native"
import relativeTime from "dayjs/plugin/relativeTime"
import React, {useEffect, useRef, useState} from "react";
import {AppRegistry} from 'react-native';
[...]

PR #397, opened 2025-06-09
fix(sentry): include standalone sentry cli in sentry plugin

mym0404:
It seems we don't need to require `@sentry/cli` as `peerDependencies`.

`@hot-updater/plugin-sentry` should be integrated as dev dependencies, hence  we can safely import it to the our `dependencies`.

This is likely to resolve the problem user uses outdated sentry cli. 

related: #341

### DIFF
--- a/plugins/sentry-plugin/package.json
+++ b/plugins/sentry-plugin/package.json
@@ -43,13 +43,10 @@
   },
   "dependencies": {
     "@hot-updater/core": "workspace:*",
-    "@hot-updater/plugin-core": "workspace:*"
+    "@hot-updater/plugin-core": "workspace:*",
+    "@sentry/cli": "^2.46.0"
   },
   "devDependencies": {
-    "@sentry/cli": "^2.43.0",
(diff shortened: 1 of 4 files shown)


plugins/sentry-plugin/package.json
```diff
@@ -43,13 +43,10 @@
   },
   "dependencies": {
     "@hot-updater/core": "workspace:*",
-    "@hot-updater/plugin-core": "workspace:*"
+    "@hot-updater/plugin-core": "workspace:*",
+    "@sentry/cli": "^2.46.0"
```
gronxb:
```suggestion
    "@sentry/cli": "2.46.0"
```

gronxb: Since it’s being brought in as a dependency, it would be better to lock the version.

gronxb (approved):
👍

plugins/sentry-plugin/package.json
mym0404: sure,

PR #409, opened 2025-06-11
fix(firebase): index default

### LINKED ISSUE #407: Firebase Cannot read properties of undefined (reading 'concat'). Error while creating cloud function
### Precheck

0.18.5

### Version of react-native

0.74.5

### Platforms

Android, iOS

### Architecture

New Architecture(Fabric)

Firebase

### Expo App

- [ ] I am using Expo

### Bundler

metro

### What happened?

Hi,

While selecting a region for my cloud function, an error about Firestore indexes is thrown.

Here is the [location](https://github.com/gronxb/hot-updater/blob/main/plugins/firebase/iac/index.ts) where the error happening.

We are trying to get actual indexes like this :
`originalIndexes = originalStdout ?? []`

While if not found it should be this (at least for my case)
`{ indexes: [], fieldOverrides: [] }`

Hence the error `Cannot read properties of undefined (reading 'concat')`

### Relevant log output

```shell
◇  Select Region
◇  Using new functions in region: europe-west1
file:///Users/damien/Work/project/node_modules/@hot-updater/firebase/dist/iac/index.js:8191
	const mergedIndexes = originalIndexes.indexes.concat(newIndexes.indexes);
	                                              ^

TypeError: Cannot read properties of undefined (reading 'concat')
    at mergeIndexes (file:///Users/damien/Work/project/node_modules/@hot-updater/firebase/dist/iac/index.js:8191:48)
    at deployFirestore (file:///Users/damien/Work/project/node_modules/@hot-updater/firebase/dist/iac/index.js:8209:24)
    at async Module.runInit (file:///Users/damien/Work/project/node_modules/@hot-updater/firebase/dist/iac/index.js:8354:2)
    at async Command.init (file:///Users/damien/Work/project/node_modules/hot-updater/dist/index.js:21019:4)

Node.js v20.12.0
```

_No response_

### Code of Conduct

jingjing2222:
## Before
There is not default value of fieldoverrides.
It occurred concat error

## after 

```typescript
originalIndexes = originalStdout ?? { indexes: [], fieldOverrides: [] };
```

and add exact type for index
https://firebase.google.com/docs/reference/firestore/indexes?hl=ko

closed #407

### DIFF
--- a/plugins/firebase/iac/index.ts
+++ b/plugins/firebase/iac/index.ts
@@ -77,8 +77,27 @@ interface FirebaseFunction {
 
 type FirebaseIndex = {
   collectionGroup: string;
-  queryScope: string;
-  fields: { fieldPath: string; order: string }[];
+  queryScope: "COLLECTION" | "COLLECTION_GROUP";
+  fields: Array<{
+    fieldPath: string;
+    order?: "ASCENDING" | "DESCENDING";
+    arrayConfig?: "CONTAINS";
+    vectorConfig?: {
+      dimension: number;
+      flat: Record<string, never>;
+    };
+  }>;
+};
+
+type FieldOverride = {
+  collectionGroup: string;
+  fieldPath: string;
+  indexes: Array<{
+    queryScope: "COLLECTION" | "COLLECTION_GROUP";
+    order?: "ASCENDING" | "DESCENDING";
+    arrayConfig?: "CONTAINS";
+  }>;
+  ttl?: boolean;
 };
 
 function normalizeIndex(index: FirebaseIndex) {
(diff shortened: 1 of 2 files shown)


plugins/firebase/iac/index.ts
```diff
@@ -77,8 +77,27 @@ interface FirebaseFunction {
...
+  queryScope: "COLLECTION" | "COLLECTION_GROUP";
+  fields: Array<{
+    fieldPath: string;
+    order?: "ASCENDING" | "DESCENDING";
+    arrayConfig?: "CONTAINS";
+    vectorConfig?: {
+      dimension: number;
+      flat: Record<string, never>;
+    };
+  }>;
```
gronxb:
```suggestion
  fields: {
    fieldPath: string;
    order?: "ASCENDING" | "DESCENDING";
    arrayConfig?: "CONTAINS";
    vectorConfig?: {
      dimension: number;
      flat: Record<string, never>;
    };
  }[];
```

Issue #359, opened 2025-05-30
fingerprint create: Something went wrong


khorark:
- [x] Infrastructure may have changed. Did you retry `hot-updater init`?
- [x] Is your Node version v20 or higher?
- [x] Are you aware that the update will only work properly if you `hot-updater deploy` after building the native app?

## Problem

I use react-native cli (not expo) and when I run command `npx hot-updater fingerprint create` I get an error `Something went wrong`. I think it depends by new strategy at version 0.18 for use fingerprinting and now as I see it depends by @expo/fingerprint lib. 

What should I do for fix it? Thx

## Environment
- React Native Version: 0.79.1
- `hot-updater` Version:  0.18.0
- Platform: both
- OS: macos

gronxb:
I’ll fix this quickly and deploy it right away!

gronxb:
Released in 0.18.1

Ajmal0197:
Hi, I am getting this.
    "hot-updater": "^0.19.1",

<img width="596" alt="Image" src="https://github.com/user-attachments/assets/79965aad-dfd2-4736-ab1b-ac6e32b73aa0" />

gronxb:
Hmm, since there’s no error showing up, I’m not sure what the issue is. I think I need to improve the error messages.

gronxb:
@Ajmal0197 Would you like to try it in version 0.19.2? It looks like the issue will be resolved in that version.

Ajmal0197:
> @Ajmal0197 Would you like to try it in version 0.19.2? It looks like the issue will be resolved in that version.
> 

Thanks man, issue is fixed

PR #414, opened 2025-06-13
fix: print error

jingjing2222:
issue #359 
print error for create fingerprint


packages/hot-updater/src/commands/fingerprint.ts
```diff
@@ -12,75 +12,91 @@ import picocolors from "picocolors";
 
 export const handleFingerprint = async () => {
   const s = p.spinner();
-  s.start("Generating fingerprints");
+  try {
```
gronxb:
```suggestion
```

gronxb:
Dupe #415

PR #418, opened 2025-06-13
feat: add debug option while creating fingerprint

r0b0t3d:
# Requirements
Sometime I'd like to know which files or folders causing the fingerprint changed, I need to add debug option to show it

# Solutions
Add debug option when configuring hot updater

```
export default defineConfig({
  build: bare({ enableHermes: true }),
  ...
  fingerprint: {
    debug: true,
  }
});
```

### DIFF
--- a/.changeset/flat-actors-ring.md
+++ b/.changeset/flat-actors-ring.md
@@ -1,0 +1,6 @@
+---
+"hot-updater": patch
+"@hot-updater/plugin-core": patch
+---
+
+Add debug option while creating fingerprint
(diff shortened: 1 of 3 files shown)


gronxb:
Thanks for your contribution ! ~But I think it would be better if it only works in the CLI, so I think it should only run when it’s `hot-updater fingerprint create --debug`.~

gronxb:
Ah, now I understand the purpose. The fingerprint.json file provides more detailed information. I think we can keep it—thanks!

Issue #339, opened 2025-05-16
Local images doesn't render during the session that a new OTA update is downloading


lailton-b:
- [ x ] Infrastructure may have changed. Did you retry `hot-updater init`?
- [ x ] Is your Node version v20 or higher?
- [ x ] Are you aware that the update will only work properly if you `hot-updater deploy` after building the native app?

## Problem

Local images doesn't render during the session that a new OTA update is downloading (for the issue to happen, the App must have a previous OTA update installed).

## Message

Hey! I ran into this problem recently, after some of our users started to send videos of a screen that had an infinite loading. After some good hours debugging it, I found out that I had a React Native's Image Component on that screen, rendering a local image in the property "defaultSource". For some reason, using this property with a local image, in a session that is downloading a new OTA update (when the App already has an update installed), creates a bug on react-native useStates, they simply wouldn't re-render the screen.

Once I removed the "defaultSource" attribute, the problem was solved. Then I decided to remain on that session and do a run through the whole App to see if I would find any new problems. That's when I realized that every Component that rendered a local image had it's images missing. I tested both on Android and iOS, and I also updated hot-updater to the latest version, running hot-updater init, using Node 20 and building the app again. The problem still happened every time during the scenario described above (only during the single session where the App is downloading a new update, after already having an update installed).

I proceeded to make some tests, I really wanted to bring a POC with the issue being ready to be reproduced, but I couldn't make it happen with a brand new App. I thought that the issue could be specific to our application, but I tried simulating it on another application of ours, and the issue happened!

Both of our applications are using React Native 0.71.19, but I also tried upgrading one of them to React Native 0.73.11, and the issue remained.

Doing a little more testing, I rewrote our App component, so it would render a basic screen, with just a local image at it's center (basically the same thing that I did with the brand new App), and the issue didn't happen... The image would render every time. I've been trying to bring a step by step reproduction for 3 days now and unfortunately I have other tasks to do at my job, I'm not sure when I'll be able to continue debugging this issue. This was all the information that I gathered, it seems like the issue is somehow connected to having a bunch of information to render, it doesn't seem to happen with a simple application.

I'm reporting the issue to see if someone is going through the same thing, I'll really appreciate some help!

## Steps to Reproduce

1 - Deploy a update;
2 - Download and install the update on the App;
3 - Deploy a second update;
4 - Open the App (if the update is downloading during the session, local images shouldn't be able to render);

Every other update forward will carry the same problem, as long as the App already have an update installed.

[...]

PR #431, opened 2025-06-19
fix: delete previous bundle when previous bundle access is needed

jingjing2222:
## Problem
1. The app starts up by loading the existing bundle and launches based on it.
2. The hot-updater downloads a new bundle and deletes the existing bundle.
3. Since the assets that the existing bundle was accessing have disappeared, local images are no longer displayed.

## Solution
1. Records of existing bundleIds and records of newly added bundleIds during the update process
2. In cleanupOldBundles, instead of keeping only the latest bundle as before, bundles corresponding to the above two parameters are not deleted

## to-do list
1. [x] - android
2. [x] - ios

### DIFF
--- a/.changeset/plain-roses-tickle.md
+++ b/.changeset/plain-roses-tickle.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+fix: delete previous bundle when previous bundle access is needed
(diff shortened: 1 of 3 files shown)


jingjing2222:
#339

gronxb:
bugbot run

PR #441, opened 2025-06-23
chore(cli): move `commander` to `devDependencies` and bundle it

gronxb:
Sometimes, if another package uses a different version of commander, hoisting can cause errors. To avoid this, we bundle commander directly into the CLI (via tsdown).

### DIFF
--- a/.changeset/red-worlds-lie.md
+++ b/.changeset/red-worlds-lie.md
@@ -1,0 +1,5 @@
+---
+"hot-updater": patch
+---
+
+chore(cli): bundle commander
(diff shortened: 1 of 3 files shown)


.changeset/red-worlds-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+"hot-updater": patch
+---
+
+chore(cli): bundle commander
```
gronxb: chore(cli): move commander to devDependencies and bundle it

gronxb:
```suggestion
chore(cli): move commander to devDependencies and bundle it
```

PR #444, opened 2025-06-23
Feature: Store output bundle and artifacts to the user accessible path

### LINKED ISSUE #432: Feature: Store output bundle and artifacts to the user accessible path
discussed from [here](https://github.com/gronxb/hot-updater/pull/394#discussion_r2156031993)

mym0404:
# Checklist

- [x] add `.hot-updater` into `.gitignore` during `init` and `deploy` command if needed
- [x] store bundle output from `deploy` into default output folder `.hot-updater/output/bundle`
- [x] store nativeBuild output from `build:native` into default output folder `.hot-updater/output/build`

resolve #432


packages/hot-updater/src/utils/git.ts
```diff
@@ -14,3 +16,40 @@ export const getLatestGitCommit = async (): Promise<Commit | null> => {
...
+  if (!globLines.length) {
+    return;
+  }
+  const comment = "# hot-updater";
+
+  const projectDir = cwd ?? getCwd();
+  const gitIgnorePath = path.join(projectDir, ".gitignore");
+
+  if (fs.existsSync(gitIgnorePath)) {
+    const content = fs.readFileSync(gitIgnorePath, { encoding: "utf8" });
```
gronxb: https://es-git.slash.page/reference/Repository/Methods/addIgnoreRule.html#addignorerule

how about this ?

mym0404: what a my narrow view.

mym0404: I've check that function out and decided don't use it because we cannot fully determine whether the .gitignore file is really changed and it doesn't document how duplicated lines are going.

gronxb: okay!

Issue #457, opened 2025-06-26
Bug Report: Expo Plugin Forces Fingerprint Generation Even When Using `appVersion` Strategy


yuliswe:
### Precheck

- [x] Infrastructure may have changed. Did you retry `hot-updater init`?
- [x] Is your Node version v20 or higher?
- [x] Are you aware that the update will only work properly if you `hot-updater deploy` after building the native app?

### Name, Version of hot-updater

0.19.8

### Version of react-native

0.76.9

### Platforms

iOS

### Architecture

New Architecture(Fabric)

### What provider are you seeing the problem on? (multiple)

S3 & Lambda

### Expo App

- [x] I am using Expo
- [ ] Project is managed with Expo [CNG](https://docs.expo.dev/workflow/continuous-native-generation/)(android, ios directories are included in `.gitignore`)

### Bundler

metro(hermes)

## **Summary**

The Expo plugin (`withHotUpdater`) forces fingerprint generation during `expo prebuild` even when the user has explicitly configured `updateStrategy: "appVersion"`. This causes the build to fail because `ensureFingerprintConfig()` exits with an error when the strategy is `appVersion`.

## **Steps to Reproduce**

1. Configure `hot-updater.config.ts` with `updateStrategy: "appVersion"`
2. Configure `app.config.ts`:

```
  plugins: [
    [
      '@hot-updater/react-native',
      {
        channel: Env.HOT_UPDATER_CHANNEL,
      },
    ]
}
```

3. Configure `babel.config.cjs`

```
    plugins: [
      'hot-updater/babel-plugin',
    ],
```

5. Run `expo prebuild` or any Expo command that processes config plugins
6. Observe that the build fails with an error from `ensureFingerprintConfig()`

## **Expected Behavior**
When `updateStrategy` is set to `"appVersion"`, the Expo plugin should:
- Skip fingerprint generation entirely
- Only inject the channel information into native files
- Not call any fingerprint-related functions

## **Actual Behavior**

The Expo plugin (`packages/react-native/plugin/src/withHotUpdater.ts`) **always calls fingerprint generation functions** regardless of the `updateStrategy` setting:

```ts
// === iOS: Add channel and fingerprint to Info.plist ===
modifiedConfig = withInfoPlist(modifiedConfig, async (cfg) => {
  // Generate fingerprints asynchronously - ALWAYS CALLED
  const { ios } = await getFingerprint(); // This calls createAndInjectFingerprintFiles()

[...]

PR #458, opened 2025-06-27
Skip fingerprint generation when using appVersion strategy

yuliswe:
Proposal fix for #457

### DIFF
--- a/.changeset/warm-islands-pump.md
+++ b/.changeset/warm-islands-pump.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+Skip fingerprint generation when using appVersion strategy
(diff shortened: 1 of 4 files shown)


mym0404:
@gronxb could we just remove this [assertion](https://github.com/gronxb/hot-updater/blob/main/packages/hot-updater/src/utils/fingerprint/common.ts#L8-L8) or accept OP's approach?

yuliswe:
@gronxb I think you have a good point. The alternative is remove that assert, or keep that assert but catch and ignore it during build (maybe output an warning). How does that sound?

gronxb:
> catch and ignore it during build (maybe output an warning). How does that sound?

it's sound good!

yuliswe:
I'm sorry... I just changed my mind. Intentionally triggering an error then ignoring it sounds like code smell to me. I'm doubling down with the original PR. I'm not familiar with the structure of this project but it looks to me there's no harm in including node code in the react-native package because all rn build tools are on node. If you are again ok with this please merge it. Thanks :)

Issue #391, opened 2025-06-07
Native Build Management beyond OTA


mym0404:
# Overview

We decided to support deploying & building native build artifacts(`apk`, `aab`, `ipa`, ...) to become `hot-updater` a comprehensive React Native CLI solution not only OTA features.

`native:build` command will be added into program.

> [!NOTE]
Build and upload functionality that manages metadata integrated with `hot-updater` **has not been considered specifically yet**. For now, we want to prioritize the functionality to simply build artifacts.

# Checklist

- [x] #394 
- [x] #466 #464
- [x] #480
- [x] #459
- [x] #460
- [x] Connect CLI program with build functions and integrate new config options
- [ ] Upload Android native build to storage
- [ ] Upload iOS native build to storage
- [ ] Console shows native build
- [ ] Download native builds from console
- [x] Link native build with the bundle in some manner(fingerprint, channel, version, **...TBD**)

dougg0k:
Nice. 

If after you were to go further and provide what https://fastlane.tools/ does, at least the main parts of it, that would be useful. 

Like code signing and app store deploy (beta and prod).

Though many actions seem to have their own list of features (when not just one), it could be used as reference. https://docs.fastlane.tools/actions/

---

Here is a silly way to have a single source of truth for versions. That could be included. To not have to upgrade through native files. I suppose other parts like sdk versions in both platforms could be added too, simpler to manage, be through package.json or another file.

 https://freedium.cfd/https://medium.com/swlh/react-native-best-practices-single-source-of-app-version-eec818408c3

mym0404:
@dougg0k you can say that again, I like fastlane to manage native parts since few years ago, and built a wrapper utilties for expo [expo-release-it]. I'm migrating and merging features into hot-updater from that.

expo-release-it was a wrapper on the  top of fastlane but in hot-updater, I don't want to depend on only fastlane like other cli tools are doing. I'm implementing a core features incrementally.

Thank you for your suggestion.

Gautham495:
@mym0404 - Are you planning to add all of the fastlane features like directly deploying the native build to Play Console and Apple iConnect for both Testflight and production?

That will be very valuable.

dougg0k:
I never seen a fluid experience on this matter before with other online CI / automation solutions, last one being bitrise. If this were properly implemented in all details / management, without requiring any manual actions in the stores console, that would be unique.

[...]

PR #459, opened 2025-06-27
iOS Native Build

mym0404:
It's a part of plan for #391.


mym0404:
This iOS Native Build feature is now being developed together with #480 which includes comprehensive native build system improvements for both Android and iOS platforms.

Moving the iOS implementation to #480 to consolidate all native build features in a single PR for better coordination and testing.

PR #463, opened 2025-06-28
feat(datadog): Introduce Datadog plugin to upload sourcemaps

valeriobelli:
Most of what has been done here derives from the Datadog documentation https://docs.datadoghq.com/real_user_monitoring/mobile_and_tv_monitoring/react_native/error_tracking.

### DIFF
--- a/docs/docs/guide/plugins/datadog.mdx
+++ b/docs/docs/guide/plugins/datadog.mdx
@@ -1,0 +1,124 @@
+import { Tab, Tabs } from 'rspress/theme';
+import { PackageManagerTabs } from '@theme';
+import { normalizeImagePath } from 'rspress/runtime';
+
+# Datadog Source Maps Plugin
+
+The `withDatadog()` plugin for hot-updater enables automatic sourcemap upload to Datadog during the update bundle build process.
+
+This helps you track production crashes with accurate stack traces tied to your React Native source code.
+
+## Prerequisites
+
+- Datadog Account: Sign up here if you don’t have one.
+- Auth Token: Generate a token from your Datadog dashboard. This must be defined as the environment variable `DATADOG_API_KEY`.
+- Datadog host: in order to communicate with Datadog, you need to specify the host to which the Auth token refers using the environment variable `DATADOG_HOST` (e.g., `DATADOG_HOST=datadoghq.eu`)
+- Install the plugin:
+
+<PackageManagerTabs command={
+  {
+    npm: "npm install @hot-updater/datadog-plugin --save-dev",
+    pnpm: "pnpm add @hot-updater/datadog-plugin -D",
+    yarn: "yarn add @hot-updater/datadog-plugin -D",
+  }
+} />
+
+::: tip Stable execution
+`withDatadog` internally uses `@datadog/datadog-ci` executed using `npx`. Although not required, you can install it as a `devDependency` to avoid unexpected HTTP calls failing during execution (e.g., in a CI environment). 
+:::
+
+## Step 1: Wrap Your Build Plugin
+
+Use `withDatadog()` to wrap any compatible build plugin such as metro.
+
+Once wrapped, sourcemaps will automatically be uploaded to Datadog when running the `hot-updater deploy` process.
+
+<Tabs>
+  <Tab label="Metro">
+
+```ts title="hot-updater.config.ts"
+import { bare } from "@hot-updater/bare";
+import { withDatadog } from "@hot-updater/datadog-plugin";
+import { defineConfig } from "hot-updater";
+import "dotenv/config";
+
+export default defineConfig({
+  build: withDatadog(
+    bare({
+      enableHermes: false, // or true, depending if you want to use it
+      sourcemap: true, // [!code ++] Required for sourcemap upload
+    }),
+    {
+      buildNumber: "build number",       // [!code ++] The build number through which Datadog will sort the uploaded sourcemaps belonging to the same version
+      releaseVersion: "release version",    // [!code ++] This is the release version that must match the release version used to stream logs from the app
+      service: "service", // [!code ++] The datadog service to which the sourcemaps belong
+    },
+  ),
+  // .. your other config
+});
+```
+</Tab>
+<Tab label="Expo">
+
+```ts title="hot-updater.config.ts"
+import { withDatadog } from "@hot-updater/datadog-plugin";
+import { defineConfig } from "hot-updater";
+import { expo } from "@hot-updater/expo";
+
+export default defineConfig({
+  build: withDatadog(
+    expo({
+      sourcemap: true, // [!code ++] Required for sourcemap upload
+    }),
+    {
+      buildNumber: "build number",       // [!code ++] The build number through which Datadog will sort the uploaded sourcemaps belonging to the same version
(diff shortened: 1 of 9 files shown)


gronxb:
Are there any use cases where withSentry and withDatadog are used together, just out of curiosity?

valeriobelli:
> Are there any use cases where withSentry and withDatadog are used together, just out of curiosity?

Hey @gronxb 👋🏽 
The only case that came to my mind is when using Datadog for RUM and Sentry for error tracking. I'm not sure if it's a typical use case, though. What are your thoughts on this matter?

gronxb:
So in hot-updater.config.ts, we need to use HoC in a nested way like this, right?
```tsx
build: withDatadog(
              withSentry(
                   bare({enabledHermes: true}),
              )
          ),
```

It looks like HoC nesting works as shown above. But when used together, it could easily turn into a nesting hell. So I’ll probably need to prepare an interface like the one below in the near future.

```tsx
combinePlugins([
    withSentry({
      org: "your-org-slug",      
      project: "your-project",   
      authToken: process.env.SENTRY_AUTH_TOKEN!,
    }),
    withDatadog({
     ...
     }),
    bare({
        enableHermes: true,
    }),
])
```

docs/docs/guide/plugins/datadog.mdx
```diff
@@ -0,0 +1,124 @@
...
+This helps you track production crashes with accurate stack traces tied to your React Native source code.
+
+## Prerequisites
+
+- Datadog Account: Sign up here if you don’t have one.
+- Auth Token: Generate a token from your Datadog dashboard. This must be defined as the environment variable `DATADOG_API_KEY`.
+- Datadog host: in order to communicate with Datadog, you need to specify the host to which the Auth token refers using the environment variable `DATADOG_HOST` (e.g., `DATADOG_HOST=datadoghq.eu`)
+- Install the plugin:
+
+<PackageManagerTabs command={
```
gronxb, line 18: `datadog-ci`

Should I also instruct them to install this?

plugins/datadog-plugin/src/withDatadog.ts
```diff
@@ -0,0 +1,131 @@
...
+        await fs.promises.copyFile(
+          selectedBundleFilePath,
+          tmpDirBundleFilePath,
+        );
+        await fs.promises.copyFile(
+          selectedSourcemapFilePath,
+          tmpDirBundleSourcemapFilePath,
+        );
+
+        await execa("npx", [
```
gronxb: How about checking the version of datadog-ci and throwing an error to prompt installation if it’s not installed?

docs/docs/guide/plugins/datadog.mdx
```diff
@@ -0,0 +1,124 @@
...
+  }
+} />
+
+::: tip Stable execution
+`withDatadog` internally uses `@datadog/datadog-ci` executed using `npx`. Although not required, you can install it as a `devDependency` to avoid unexpected HTTP calls failing during execution (e.g., in a CI environment). 
+:::
+
+## Step 1: Wrap Your Build Plugin
+
+Use `withDatadog()` to wrap any compatible build plugin such as metro.
```
gronxb:
```suggestion
Use `withDatadog()` to wrap any compatible build plugin such as bare.
```

docs/docs/guide/plugins/datadog.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+Once wrapped, sourcemaps will automatically be uploaded to Datadog when running the `hot-updater deploy` process.
+
+<Tabs>
+  <Tab label="Metro">
```
gronxb:
```suggestion
  <Tab label="Bare">
```

plugins/datadog-plugin/src/withDatadog.ts
valeriobelli: That's a leftover. The current implementation is not working. 😬 

We have two ways. The first is to declare `npx @datadog/datadog-ci`, which removes the need for static installation in most cases. However, in my experience with CI environments, this extra HTTP call could indeed fail, which brings us to the [tip](https://github.com/gronxb/hot-updater/pull/463/files#diff-c68ca426eb1edcb8510a5758b4830141c2b228a74ee92923667270b497b9a440R26-R28) mentioned in the docs. Therefore, an alternative approach is to declare `@datadog/datadog-ci` as a peer dependency by retaining the current `npx` declaration and ensuring stable execution, as the installation of that dependency is required. 

WDYT?

gronxb: yes. i think better option `peerDeps`

valeriobelli:
> So in hot-updater.config.ts, we need to use HoC in a nested way like this, right?
> 
> ```tsx
> build: withDatadog(
>               withSentry(
>                    bare({enabledHermes: true}),
>               )
>           ),
> ```
> 
> It looks like HoC nesting works as shown above. But when used together, it could easily turn into a nesting hell. So I’ll probably need to prepare an interface like the one below in the near future.
> 
> ```tsx
> combinePlugins([
>     withSentry({
>       org: "your-org-slug",      
>       project: "your-project",   
>       authToken: process.env.SENTRY_AUTH_TOKEN!,
>     }),
>     withDatadog({
>      ...
>      }),
>     bare({
>         enableHermes: true,
>     }),
> ])
> ```

Yes, it might happen (still, I'm not sure how often). Unless there are already other planned use cases for injecting multiple plugins during build steps, I'd personally wait and encourage the community to provide feedback before introducing more APIs. WDYT?

gronxb:
Yes, that’s right. I’m planning to think about this later after merging it.
By the way, is there any test evidence for this?

valeriobelli:
> Yes, that’s right. I’m planning to think about this later after merging it. By the way, is there any test evidence for this?

Setting it all up from scratch in a separate environment may require some effort. So, I don't know if it's enough for you, but these screenshots come from the actual implementation and usage in a stage environment as early evidences

upload-android

upload-ios

uploaded-artefact

log-sourcemaps-2

log-sourcemaps


I see a couple of critical issues:
- We obtain absolute paths during sourcemap generation, which are, of course, different between the CI environment and the actual repository structure. For clarity, my codebase is a monorepo. Currently, the result of this is what you can see in the last screenshot in the tooltip, and the fact that Datadog can't link logs to repository files. I think we should use the RN option `--sourcemap-sources-root` that `react-native bundle` command exposes. Considering the `bare` build plugin, I was thinking something like
```typescript
export default {
	build: bare({
		enableHermes: true,
		sourcemap: { 
			sourcesRoot: '<custom root>',
		}, // We can also keep `boolean` as an available type for backward compatibility
	})
}
```
- We can see that we can still see missing sourcemaps due to Hermes `InternalBytecode.js` file, but I don't think we can do much here. I can explore the possibilities here separately.



In my opinion, these two issues are not strictly related to this PR. What do you think about fixing them in a separate PR?

gronxb:
I’ll take a closer look at what you said when I have some time.
It sounds like a good idea to separate the PRs.
Thanks for your contribution!

gronxb (approved):

gronxb:
https://www.npmjs.com/package/@hot-updater/datadog-plugin 🚀

PR #466, opened 2025-06-29
Android Injectable minBundleId

### LINKED ISSUE #464: Feature: Android Injectable minBundleId on native build
## Overview

I thought `build:native` should generate a minBundleId and inject it into the build for compatibility with existing APIs.

In Android, `minBundleId` has been calculated on build time with `BuildConfig` in `build.gradle` (thank you for already flexible configuration easy to use [gradle cmdline project property](https://docs.gradle.org/current/userguide/build_environment.html#setting_a_project_property)).

But in iOS, it is calculated based on preprocessor constant `__DATE__` and `__TIME__` making hard to inject. Although, I guess we can tweak this tiny logic with easy way.

## Spec of generated `minBundleId` from `build:native` command

- ephemeral: the value shouldn't be remained after build. Because user can use other native build system at the future.
- cmdline injection in `build:native` step.
- one-time configuration pipeline to pass cmdline `minBundleId` to native runtime. (I'm figuring out this)

## Related

- #429

mym0404:
close #464 

# Test Plan

## Android

image


mym0404:
@gronxb I want to split PR because iOS part requires #459 first.

p.s.) you can ease your dev with fixed `pnpm build:dev` command.

gronxb:
@mym0404 okay.

packages/react-native/android/src/main/java/com/hotupdater/HotUpdaterImpl.kt
```diff
@@ -69,10 +69,16 @@ class HotUpdaterImpl(
     }
 
     /**
-     * Generates a bundle ID based on build timestamp
+     * Get minimum bundle ID string
```
mym0404, line 72:
```suggestion
     * Get minimum bundle ID
```

packages/hot-updater/src/utils/nativeBuild/runAndroidNativeBuild.ts
```diff
@@ -12,8 +13,10 @@ export const runAndroidNativeBuild = async ({
 }): Promise<{ buildDirectory: string; outputFile: string }> => {
   const androidProjectPath = path.join(getCwd(), "android");
 
+  const bundleId = uuidv7();
```
gronxb: minBundleId
```suggestion
  function generateMinBundleId() {
  const timestamp = BigInt(Date.now());

  // UUIDv7: 48-bit timestamp in milliseconds since epoch
  const timeHigh = Number((timestamp >> 16n) & 0xFFFFFFFFn);
  const timeLow = Number(timestamp & 0xFFFFn);

  // 4-bit version (0111 for version 7)
  const version = 0x7;

  // 2-bit variant (10xx)
  const variantBits = 0b10;

  // Random bits zero
  const randomA = 0;
  const randomB = 0;

  // Build each component
  const timeHighStr = timeHigh.toString(16).padStart(8, "0");
  const timeLowStr = timeLow.toString(16).padStart(4, "0");

  // For version field: high nibble is version, low 12 bits random (here zero)
  const versionAndRandomStr = ((version << 12) | (randomA & 0x0fff))
    .toString(16)
    .padStart(4, "0");

  // For variant field: 2 bits variant + 14 bits random (here zero)
  const variantAndRandomStr = (
    (variantBits << 14) | (randomB & 0x3fff)
  )
    .toString(16)
    .padStart(4, "0");

  // Last 3 bytes (all zeros here)
  const nodeStr = "000000000000";

  // Compose UUID string
  return (
    `${timeHighStr}-${timeLowStr}-${versionAndRandomStr}-` +
    `${variantAndRandomStr}-${nodeStr}`
  );
}

    const minBundleId = generateMinBundleId();
```
Hot about this?

gronxb: image

For reference, in the case of minBundleId, we fill in the timestamp part at the beginning, but set the remaining random area to 0.
If we make the random area completely random as well, it becomes indistinguishable from the IDs of regular OTA bundles.
So, by setting that part to 0, we treat it as the minBundleId, and consider it an identifier for the built-in bundle.

gronxb (changes requested):

packages/hot-updater/src/utils/nativeBuild/runAndroidNativeBuild.ts
gronxb: IMG_4102


https://generateuuid.online/converter/v7-to-timestamp

this util compat uuidv7

mym0404: Thank you for suggestion. looks good!

gronxb (approved):

packages/hot-updater/src/utils/nativeBuild/runAndroidNativeBuild.ts
mym0404: @gronxb I've done.

PR #467, opened 2025-06-30
fix(fingerprint): skip gitignore and package json scripts

gronxb:
https://docs.expo.dev/versions/latest/sdk/fingerprint/#enums

### DIFF
--- a/.changeset/fair-dots-sell.md
+++ b/.changeset/fair-dots-sell.md
@@ -1,0 +1,5 @@
+---
+"hot-updater": patch
+---
+
+skip gitignore and package json scripts


packages/hot-updater/src/utils/fingerprint/common.ts
```diff
@@ -27,6 +31,7 @@ export function getFingerprintOptions(
       "**/.gitignore",
```
mym0404, line 31: could we delete this?

mym0404 (approved):

packages/hot-updater/src/utils/fingerprint/common.ts
```diff
@@ -27,6 +31,7 @@ export function getFingerprintOptions(
...
       ...options.ignorePaths,
     ],
+    sourceSkips: SourceSkips.GitIgnore | SourceSkips.PackageJsonScriptsAll,
```
gronxb, line 34: If you add the flag like this, that file will be skipped.

@mym0404 

same feature !

packages/hot-updater/src/utils/fingerprint/common.ts
gronxb: https://docs.expo.dev/versions/latest/sdk/fingerprint/#gitignore

packages/hot-updater/src/utils/fingerprint/common.ts
gronxb: https://docs.expo.dev/versions/latest/sdk/fingerprint/#gitignore

PR #532, opened 2025-08-19
feat: default updateStrategy change from `fingerprint` to `appVersion`

gronxb:
Fingerprint changes too frequently and requires a higher level of expertise. Therefore, for now, since it can be confusing to use as a default, we use the more familiar appVersion as the default.


gronxb:
@claude

gronxb:
@claude

PR #543, opened 2025-08-23
feat: use pnpm catalog for workspace dependencies

mym0404:
Manage versions of `picocolors`, `execa`, `@types/node` with catalog.

This codes are just leftovers from the other branch.


mym0404:
@gronxb  Do you expect any problems presented by matched versions of these packages?

PR #547, opened 2025-08-26
Fix/aws invalidate target app version semver

### LINKED ISSUE #541: check-update returns null when minBundleId === bundleId
### Precheck

0.20.5

### Version of react-native

0.79.6

### Platforms

iOS

### Architecture

New Architecture(Fabric)

S3 & Lambda

### Expo App

- [ ] I am using Expo

### Bundler

metro (hermes)

### What happened?

So everything init'd just fine, native build succeeded and `HotReload` module reports proper app version/channel/minBundleId/bundleId. `deploy` also works, all the values match, the new bundleId is also newer than the previous. App, however, fails to fetch that update

First I tried testing manually, GET `check-update/app-version/ios/3.0.0/development/0198d2fe-f508-7000-8000-000000000000/0198d2fe-f508-7000-8000-000000000000` (both pulled from the app by calling `HotUpdater.getBundleId()` and `HotUpdater.getMinBundleId()` respectively) returns `null`, not some sort of JSON with a message, just `null`

Double checked the app version/channel and ids, all correct on both app and console (deployed using `hot-updater deploy -p "ios" -t "3.0.x" -c "development" -m "TEST"`), in console the bundleId is `0198d302-15dc-7bc3-865e-a51d444bbf56`

Tried tinkering with IDs - changing either minBundleId or bundleId in the URL results into proper response, telling me to update (with both valid and invalid IDs), as long as they don't match the API seems to respond properly

One more thing I noticed: rebuilding the app again doesn't change the bundleId. Clearing DerivedData obviously refreshes it but I wouldn't expect this behavior. I cannot confidently say that we don't do some aggressive caching with DerivedData in our pipelines to save some build time, might need to check later

Did I do something wrong that resulted into IDs matching which shouldn't happen?

Thanks in advance!

### Relevant log output

```shell

```

_No response_

### Code of Conduct

gronxb:
Fixed #541

### DIFF
--- a/.changeset/hip-actors-work.md
+++ b/.changeset/hip-actors-work.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/plugin-core": patch
+---
+
+fix(aws): widen invalidation scope when targetAppVersion covers a broader range
--- a/plugins/plugin-core/package.json
+++ b/plugins/plugin-core/package.json
@@ -50,11 +50,13 @@
     "cosmiconfig": "9.0.0",
     "cosmiconfig-typescript-loader": "5.0.0",
     "fast-glob": "3.3.3",
-    "oxc-transform": "0.82.1"
+    "oxc-transform": "0.82.1",
+    "semver": "^7.7.2"
   },
   "devDependencies": {
     "@hot-updater/plugin-core": "workspace:*",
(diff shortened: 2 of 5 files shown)


gronxb:
@claude

Issue #533, opened 2025-08-19
RFC: Dual-Part Fingerprint for Native Build and OTA Updates


gronxb:
## Background

Currently, a single fingerprint is used to represent both native build identity and OTA update compatibility.
This approach is problematic because the fingerprint changes too frequently, making it unsuitable for OTA stability and complicating caching strategies.

## Proposal

Introduce a dual-part fingerprint model that separates concerns between compatibility and caching:
* nativeFingerprint (160bit)
Full fingerprint consisting of two segments:
* compat part (80bit) = otaFingerprint
* caching part (80bit)
Provides a complete identity for native build isolation and caching
* Changes frequently, tied directly to build artifacts
* otaFingerprint (80bit)
* Identical to the compat part (front 80bit) of nativeFingerprint
* Represents OTA compatibility groups
* Changes less frequently, focusing only on major compatibility factors (e.g., appVersion, ABI, RN minor version, Native Module Packages)

## Mapping
* nativeFingerprint = compat part (otaFingerprint, 80bit) + caching part (80bit)
* **Thus**:
* Multiple nativeFingerprints can share the same otaFingerprint if their compat part is identical

**This structure allows**:
* Native builds to maintain fine-grained cache control via the caching part
* OTA updates to remain stable by referencing only the compat/otaFingerprint

**Compatibility Strategy**
* OTA compatibility is determined only by the otaFingerprint (compat part)
* Within the same otaFingerprint, multiple nativeFingerprints can exist (differing in caching part) but are still OTA-compatible
* This ensures OTA updates are not blocked by minor changes in the caching part

**Benefits**
* Stability: OTA updates rely only on the compat/otaFingerprint, avoiding unnecessary update rejections
* Precision: Native builds use the full 160bit nativeFingerprint, ensuring strict caching and isolation
* Rollback Simplicity: Any two nativeFingerprints sharing the same otaFingerprint are rollback-compatible
* Flexibility: The caching part can encode additional metadata such as platform folder changes, app.json variations, or build-time noise

**Next Steps**
1. Define the exact set of inputs included in:
* otaFingerprint (compat part)
* caching part

2. Implement fingerprint calculation logic to split into two 80bit segments

3. Update server/client matching logic:
* OTA candidates matched by otaFingerprint only
* Native caching keyed by full nativeFingerprint

gronxb:
[...]

PR #551, opened 2025-08-26
feat(fingerprint): calculate OTA fingerprint only in native module

### LINKED ISSUE #477: Fingetprint mismatch everytime i make release build from Android studio
I verified setup and ios runs well when i run iOS app in simulator with Release scheme, It works smoothly,

Now when i come to Android part, when i make APK to install on device every time i gets `Android fingerprint mismatch`, the every single time even without any native code change. 

Image

And then i tried with `npx react-native run-android --mode Release` command, It works fine, fingerprint matched and uploaded bundle and got tested well. 
What is the workaround for Android studio?

gronxb:
Milstone #533 

Fixed #477


gronxb:
@claude

gronxb:
@claude

PR #562, opened 2025-09-05
feat(fingerprint): bring back ignorePaths

r0b0t3d:
I was using version 0.19.9 with option `ignorePaths` for fingerprint. It was working well. But when upgrading to 0.20, that option was gone.

This PR bring it back so that you can pass some paths that you'd like to ignore when generating fingerprint

### DIFF
--- a/.changeset/new-toes-beam.md
+++ b/.changeset/new-toes-beam.md
@@ -1,0 +1,6 @@
+---
+"hot-updater": patch
+"@hot-updater/plugin-core": patch
+---
+
+feat(fingerprint): bring back ignorePaths


gronxb:
Thanks for the contribution! In 0.20, I thought ignorePaths wouldn’t be necessary anymore since the calculation method was changed to use a whitelist that ignores all files by default and only allows native code files. But what kind of files should be additionally ignored? Could you give me an example?

r0b0t3d:
@gronxb  My experiences 
1. Create fingerprint
2. Build android & ios app
3. Check fingerprint -> mismatch
4. Create fingerprint again
```
❯ yarn hot-updater fingerprint create
yarn run v1.22.22
$ /Volumes/External/Workspace/klassapp/source/edular/edular-mobile/node_modules/.bin/hot-updater fingerprint create
│
◇  Created fingerprint.json
│
●  Changed Android paths:
│
●    android/app/src/main/res/values/strings.xml
│
●  Changed iOS paths:
│
●    ios/OneSignalNotificationServiceExtension/Info.plist
│
●    ios/edular/Info.plist
│
◆  fingerprint.json has changed, you need to rebuild the native app.
│
●  iOS Fingerprint Changes:
│
●    Changed: node_modules/react-native-config
│
●  Android Fingerprint Changes:
│
●    Changed: node_modules/react-native-config
```
Here is the file causing the change `node_modules/react-native-config/ios/ReactNativeConfig/GeneratedDotEnv.m`
image

So I just want to ignore some files in some specific cases

gronxb (approved):

gronxb:
Released in 0.20.11
Thanks !

PR #615, opened 2025-10-24
feat: validate bundle  file with fileHash

### LINKED ISSUE #614: Incomplete Bundle Downloads Lead to JS Corruption
### Precheck

0.20.13

### Version of react-native

0.81.4

### Platforms

Android, iOS

### Architecture

New Architecture(Fabric)

S3 & Lambda

### Expo App

- [ ] I am using Expo

### Bundler

metro(hermes)

### What happened?

Hello! We started noticing a problem with `Crashlytics` that grew rapidly after we uploaded the first versions using `hot-updater`, mainly on Android:

`com.facebook.jni.CppException - Compiling JS failed: Wrong bytecode version. Expected 89 but got 84 Buffer size 9064734 starts with: c61fbc03c103191f540000003223e854 and has protection mode(s): r--p`

or 

`Fatal Exception: com.facebook.jni.CppException: Compiling JS failed: 1:1:unrecognized Unicode character \u0000 Buffer size 13986743 starts with: 00000000000000000000000000000000 and has protection mode(s): r--p`

This error typically occurs when React Native attempts to execute a corrupted bundle. We addressed this by applying a patch where we are now sending a `.txt` file containing an `MD5 checksum` (hash) inside the `bundle.zip`, generated from the `index.android.bundle`. When the bundle is downloaded, we generate a new hash and compare it with the hash in the `.txt file`. If the hashes match, the file is not corrupted (ideally, this hash should be received in the `api/check-update` request, but that adjustment seemed too complex for us). We recommend that you implement something similar, if possible—it would be great!

Upon analyzing the CloudWatch logs more closely, we noticed a possible cause: we observed that many `bundle.zip` files are being downloaded only partially. Below are some bundle sizes and the number of requests:

5.7764MB - 295.000 downloads
5.776489MB - 56.739 downloads
0.9844MB - 485 downloads

And this is just the start, donwloads from 1 hour ago. It seems likely that in some cases on Android, a Socket error is not being returned, and the corrupted JavaScript problem ends up occurring (especially in downloads where only a few bytes were not downloaded, as the `unzip` process might still succeed in those situations). 
This also impacts AWS costs, as users need to download the same file multiple times, starting from the beginning.

Given this, do you think it would be prudent to use something like the Android DownloadManager? It already handles internet connectivity issues, pausing or retry downloads, in addition to having better performance when the application is running in the background.

### Relevant log output

```shell
Fatal Exception: com.facebook.jni.CppException: Compiling JS failed: 1:1:unrecognized Unicode character \u0000 Buffer size 13986743 starts with: 00000000000000000000000000000000 and has protection mode(s): r--p
       at com.facebook.jni.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:938)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
       at android.os.Looper.loop(Looper.java:250)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:228)
       at java.lang.Thread.run(Thread.java:923)
```

_No response_

### Code of Conduct

gronxb:
Fix #614

### DIFF
--- a/docs/docs/guide/hot-updater/updateBundle.mdx
+++ b/docs/docs/guide/hot-updater/updateBundle.mdx
@@ -30,22 +30,25 @@ async function applyAppUpdate(updateInfo: UpdateInfo) {
 
     /**
      * You can apply updates using one of two methods:
-     * 
-     * Method 1: Use the updateBundle() method from the updateInfo object
+     *
+     * Method 1: (RECOMMENDED) Use the updateBundle() method from the updateInfo object
      * - A convenience method built into the return value from checkForUpdate
-     * - Performs the same function as HotUpdater.updateBundle with all required arguments pre-filled
+     * - Automatically includes fileHash for secure verification
+     * - All required arguments are pre-filled from the checkForUpdate response
      */
-    // await updateInfo.updateBundle();
+    await updateInfo.updateBundle();
 
     /**
      * Method 2: Call HotUpdater.updateBundle() directly
      * - Explicitly pass the necessary values extracted from updateInfo
+     * - You must manually include fileHash for security
      */
-    await HotUpdater.updateBundle({
-      bundleId: updateInfo.id,
-      fileUrl: updateInfo.fileUrl,
-      status: updateInfo.status,
-    });
+    // await HotUpdater.updateBundle({
+    //   bundleId: updateInfo.id,
+    //   fileUrl: updateInfo.fileUrl,
+    //   fileHash: updateInfo.fileHash,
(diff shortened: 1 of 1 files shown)


gronxb:
@claude

gronxb:
@claude

PR #636, opened 2025-10-29
feat: Add disabled option to deploy command

tiagohs:
Hello! We used CodePush for many years, and one of the features we used the most was `--disabled`, a parameter in the CI/CD that sent an update already disabled in a channel.

This ensured that incorrect or imprudent uploads would not be immediately available to users.

I brought this to the Library through a patch, and since it might be useful for others, I decided to open a PR with it. If it makes sense, it would be great to make it available!

I'm not sure if there are any unit tests to cover this flow; I can't do that here.


gronxb (approved):
Thanks!

PR #709, opened 2025-11-23
feat: add bundle signing for cryptographic OTA verification

abdoolly:
## Summary

This PR adds optional cryptographic bundle signing to Hot Updater, enabling end-to-end verification of OTA updates. When enabled, bundles are signed during deployment using RSA-SHA256 and verified on the native side before installation.

**Security Benefits:**
- Prevents man-in-the-middle attacks on OTA updates
- Ensures bundle integrity (tamper detection)
- Restricts deployments to authorized keys only

### New CLI Commands (`packages/hot-updater`)

| Command | Description |
|---------|-------------|
| `keys generate` | Generates RSA key pairs (2048 or 4096-bit) |
| `keys export-public` | Exports public key to iOS Info.plist and Android strings.xml |
| `keys remove` | Removes public keys from native configuration files |

### Modified Commands

- **`deploy`**: Now validates signing configuration before deployment and signs bundles when enabled

### Core Package (`packages/core`)

- **New**: `signedHashUtils.ts` - Utilities for signed hash format (`sig:<signature>;sha256:<hash>`)
  - `createSignedFileHash()` - Creates signed hash format
  - `parseFileHash()` - Parses both signed and legacy formats
  - `isSignedFileHash()` - Detects signed format

### Native Implementation

**iOS (Swift):**
- New `SignatureVerifier.swift` - RSA-SHA256 verification using Security framework
- Modified `BundleFileStorageService.swift` - Integrated signature verification before bundle installation

**Android (Kotlin):**
- New `SignatureVerifier.kt` - RSA-SHA256 verification using Java Signature API
- Modified `BundleFileStorageService.kt` - Integrated signature verification

### React Native JavaScript

- **`wrap.tsx`**: New `onSignatureVerificationFailure` callback for security event handling
- **`types.ts`**: New `isSignatureVerificationError()` helper function
- **`checkForUpdate.ts`**: Handles signature field from update responses

### Signing Utilities (`packages/hot-updater/src/utils/signing/`)

| File | Purpose |
|------|---------|
| `keyGeneration.ts` | RSA key pair generation and management |
| `bundleSigning.ts` | Bundle signing with RSA-SHA256 |
| `validateSigningConfig.ts` | Configuration validation with error/warning detection |

### Configuration Parsers

- **`androidParser.ts`**: Read/write `hot_updater_public_key` in `res/values/strings.xml`
- **`iosParser.ts`**: Read/write `HOT_UPDATER_PUBLIC_KEY` in `Info.plist`

### Documentation

- **New guide**: `/docs/guides/bundle-signing.mdx` - Comprehensive guide with setup instructions, security best practices, CI/CD integration, and troubleshooting

## Configuration

```typescript
// hot-updater.config.ts
export default defineConfig({
  // ... other config
  signing: {
    enabled: true,
    privateKeyPath: "./keys/private-key.pem",
  },
});
```

## FileHash Format

Signed bundles use an embedded signature format:
```
sig:<base64_signature>;sha256:<hex_hash>
```

Unsigned bundles retain the plain SHA256 hash for backward compatibility.

## Breaking Changes

**None.** Fully backward compatible:
- Signature field is optional throughout the codebase
- Bundles without signatures work when no public key is configured in native apps
- Existing deployments continue to function without changes

## Configuration State Matrix

| Config `signing.enabled` | Native Public Key | Behavior |
|--------------------------|-------------------|----------|
| `false` | Not present | ✅ Normal updates (no verification) |
| `false` | Present | ❌ Updates rejected - app expects signatures |
| `true` | Present | ✅ Signature verified, update proceeds |
| `true` | Not present | ❌ Error: PublicKeyNotConfigured |

## Quick Start

```bash
# 1. Generate RSA key pair
npx hot-updater keys generate

# 2. Add signing config to hot-updater.config.ts
# signing: { enabled: true, privateKeyPath: "./keys/private-key.pem" }

# 3. Export public key to native apps
npx hot-updater keys export-public

# 4. Release new app version, then deploy signed bundles
npx hot-updater deploy -p ios
```

## Packages Affected

- `@hot-updater/core` (patch)
- `hot-updater` (patch)
- `@hot-updater/react-native` (patch)

### DIFF
--- a/docs/content/docs/react-native-api/wrap.mdx
+++ b/docs/content/docs/react-native-api/wrap.mdx
@@ -229,8 +229,6 @@ The `onError` callback handles all errors during the update process
 import {
   HotUpdater,
   getUpdateSource,
-  isSignatureVerificationError,
-  extractSignatureFailure,
 } from "@hot-updater/react-native";
 import { Alert } from "react-native";
 


gronxb:
Thanks for the contribution! I'm really looking forward to this feature.

There's one point I'd like to discuss — I'm trying to minimize database migrations, since they often create friction for users when upgrading.

In the current PR, a new `signature` field has been added, but I wonder if we could reuse the existing `filehash` field instead. Right now, `filehash` contains a SHA-256 hash, but if signature verification is enabled, we could store something like `sig:<signature>;sha256:<original>`.

From the native implementation's perspective, if this format doesn't match the expected hash-only value, it could just continue doing regular file hash verification. But if it detects the new format, it could trigger signature verification instead.

This way, we maintain backward compatibility while extending functionality.  
What do you think?

abdoolly:
@gronxb 
I think that would be very nice to make it easier  to use the signing feature without the friction of migration as well.
I think that will also definitely simplify the current implementation a little bit as well.
let me do that update.

Also thank you so much for your fast response appreciate that a lot.

abdoolly:
@gronxb one thing I would like to note though we can make this with this approach, but I think it's maybe worth it to add to the roadmap to create proper steps for migration with lowest friction since I think definitely we will bump into that wall of needing to add a new column at some point.

gronxb:
Yes, a new column might be added, but if we proceed with this approach, we won’t need a new column, and we currently have no plans to add one for the time being.  
What kind of things are you thinking about adding to the roadmap?

abdoolly:
That makes sense for the time being as you said.

> What kind of things are you thinking about adding to the roadmap?

I meant adding something low priority like a new cli `migrate-plugin` or a better name. 
which runs migrations and update edge functions if needed. 
That way those kind of updates are easy e.g update commands 

`pnpm add hot-updater@latest`
`npx hot-updater migrate-plugin` (checks for the active plugin from the config file and migrate the correct storage plugin)

gronxb:
Ah, I see. That’s an issue I’ve also been thinking about.  
Right now, since the hot-updater init already takes migration into account, there’s a mechanism in place that allows for migration.  
However, the important thing is that **it's very difficult to determine whether I need to migrate all the way to the server**.  
The key point is that users need to be explicitly aware that they’re expected to migrate the DB in the current situation.

I’m not sure exactly what form it will take, but I do want to provide a way for users to identify the right timing for migration and smoothly go through it.  
Right now, I’m working on laying the foundation for that.  
Thanks for the great point!

However, since there are no migration plans for the time being, it’s a minor issue for now — but I’m aware that it will eventually be necessary.

abdoolly:
@gronxb feel free to have a look now the change is updated. as discussed 
I have also updated the documentation.

packages/react-native/android/src/main/java/com/hotupdater/HotUpdaterImpl.kt
```diff
@@ -186,15 +186,17 @@ class HotUpdaterImpl(
...
      * @param fileHash SHA256 hash of the bundle file for verification (nullable)
+     * @param signature RSA-SHA256 signature of fileHash (nullable)
      * @param progressCallback Callback for download progress updates
      * @return true if the update was successful
      */
     suspend fun updateBundle(
         bundleId: String,
         fileUrl: String?,
         fileHash: String?,
+        signature: String?,
```
gronxb: How about keeping it as before by only receiving fileHash, since it’s now integrated into fileHash, and extracting the signature internally on the native side? That way, there won’t be any JS changes.

packages/core/src/signedHashUtils.ts
```diff
@@ -0,0 +1,250 @@
...
+  if (!fileHash) {
+    return null;
+  }
+
+  try {
+    return parseFileHash(fileHash);
+  } catch {
+    return null;
+  }
+}
```
gronxb: How about moving that logic to the native side instead, keeping the same behavior?

gronxb (changes requested):

gronxb:
If a native app includes a key like `hot_updater_public_key` or `HOT_UPDATER_PUBLIC_KEY`, I believe that a signature verification process must be enforced, assuming the app is capable of reading the public key.

For example, if the native app reads the `HOT_UPDATER_PUBLIC_KEY` and thus possesses the public key, but it doesn’t receive a signature from the server, that should be considered a failure.

If the app does not include a public key, there’s a possibility that the signature is only used on the server side and the verification is skipped.

Therefore, I think enforcing signature verification whenever a public key exists in the app is a safer approach.

What are your thoughts on this direction?
Or are you already doing it this way?

abdoolly:
> If a native app includes a key like `hot_updater_public_key` or `HOT_UPDATER_PUBLIC_KEY`, I believe that a signature verification process must be enforced, assuming the app is capable of reading the public key.
> 
> For example, if the native app reads the `HOT_UPDATER_PUBLIC_KEY` and thus possesses the public key, but it doesn’t receive a signature from the server, that should be considered a failure.
> 
> If the app does not include a public key, there’s a possibility that the signature is only used on the server side and the verification is skipped.
> 
> Therefore, I think enforcing signature verification whenever a public key exists in the app is a safer approach.
> 
> What are your thoughts on this direction? Or are you already doing it this way?

yes it is working that way  exactly.
As well as  I put some warning and guards so the user is knowledgable of the expected behaviour in which if he changed the configuration `signing.enabled: false` and the public keys are still there then the deployment would show him a warning that the verification will still be applied.

Also if the `signing.enabled: true` we will throw an error in deployment that the public keys are required directing the user to use the generate and export command since this is considered a mismatching configuration and should be fixed.

As explained in the configuration matrix below (this also exists in the documentation) 
image

gronxb:
your a goat 👍

gronxb:
I'm sorry. There's one more thing I'd like to discuss.

Currently, `fileHash` holds two values separated by a semicolon—`sig` and `sha256`.  
But I think signature verification alone can replace file integrity checking.  
So I was thinking, maybe we could store just one value in the DB like this:  
`fileHash=sig:<value>`

It could be either (DB Side):
- `fileHash=sig:<value>` or  
- `fileHash=<original sha256 filehash>`


Then, from the native code's point of view, if `sig:` exists, we perform signature verification;  
if not, we proceed with the traditional file hash check.

This way, we can support the feature without having to check both the signature and the file hash.  
What do you think?

packages/core/src/signedHashUtils.ts
abdoolly: While moving logic to native code seemed appealing, keeping parseFileHash in JavaScript is the better approach:

  1. Avoids code duplication - A single JS implementation vs maintaining identical logic in both iOS and Android
  2. Simpler maintenance - JavaScript is easier to read, modify, and test than native Swift/Kotlin
  4. Incomplete benefit - The JS layer still handles createSignedHash (hash + signature creation), so we can't fully eliminate JS logic anyway

  Bottom line: Native parsing adds complexity without meaningful gains since signature creation remains in JS
  regardless.

what do you think ?

gronxb: I see. But could you take a look at the link below? I think this could eliminate the need for parsing.

https://github.com/gronxb/hot-updater/pull/709#issuecomment-3569320522

abdoolly: that's actually a good point.
the parsing is then removed and simplified to be just to check if "sig:" prefix exists.
then we can move to native since it's just a simple check.

abdoolly:
> I'm sorry. There's one more thing I'd like to discuss.
> 
> Currently, `fileHash` holds two values separated by a semicolon—`sig` and `sha256`. But I think signature verification alone can replace file integrity checking. So I was thinking, maybe we could store just one value in the DB like this: `fileHash=sig:<value>`
> 
> It could be either (DB Side):
> 
> * `fileHash=sig:<value>` or
> * `fileHash=<original sha256 filehash>`
> 
> Then, from the native code's point of view, if `sig:` exists, we perform signature verification; if not, we proceed with the traditional file hash check.
> 
> This way, we can support the feature without having to check both the signature and the file hash. What do you think?

This will make a significant simplification actually.
I am updating now to
- remove parsing logic overall we will now do a simple prefix check in the native sides
- updating the fileHash format 

I will stick with the same configuration matrix mentioned above [here](https://github.com/gronxb/hot-updater/pull/709#issuecomment-3568712545)

abdoolly:
For documentation purposes this is the expected flow

```

  fileHash received
      │
      ├─ null/empty?
      │   │
      │   ├─ Public key configured? → REJECT (signing enabled, no hash)
      │   │
      │   └─ No public key → REJECT (no hash to verify)
      │
      └─ Has value
          │
          ├─ startsWith("sig:") → SIGNED MODE
          │   ├─ Extract signature = fileHash.substring(4)
          │   ├─ Calculate SHA256 of downloaded file
          │   ├─ Public key configured?
          │   │   ├─ NO → REJECT (signed bundle, can't verify)
          │   │   └─ YES → Verify signature
          │   │       ├─ VALID → Install
          │   │       └─ INVALID → REJECT
          │
          └─ Plain hash → UNSIGNED MODE
              ├─ Calculate SHA256 of downloaded file
              └─ Match? → Install / REJECT
```

gronxb:
That's Cool !!

abdoolly:
@gronxb all changes are in place.

docs/content/docs/react-native-api/wrap.mdx
```diff
@@ -221,6 +221,54 @@ export default HotUpdater.wrap({
...
+    Alert.alert(
+      "Security Warning",
+      "Update verification failed. Please update the app from the app store.",
+      [{ text: "OK" }]
+    );
+  },
+  onError: (error) => {
+    // This is also called after onSignatureVerificationFailure
+    console.error("Update error:", error);
+  },
```
gronxb: I’m wondering if it might be possible to handle everything with just onError without needing onSignatureVerificationFailure. What would be a use case for having it separately?

packages/core/src/types.ts
```diff
@@ -104,8 +104,10 @@ export interface UpdateInfo {
 export interface AppUpdateInfo extends Omit<UpdateInfo, "storageUri"> {
   fileUrl: string | null;
   /**
-   * SHA256 hash of the bundle file for integrity verification.
-   * If provided, the client will verify the downloaded file's hash before extraction.
+   * SHA256 hash of the bundle file, optionally with embedded signature.
+   * Format when signed: "sig:<base64_signature>;sha256:<hex_hash>"
```
gronxb:
```suggestion
   * Format when signed: "sig:<base64_signature>"
```

packages/react-native/ios/HotUpdater/Internal/HotUpdaterImpl.swift
```diff
@@ -145,13 +145,16 @@ import React
             // Extract fileHash if provided
             let fileHash = data["fileHash"] as? String
 
+            // Extract signature if provided
+            let signature = data["signature"] as? String
+
```
gronxb: There’s no need for it anymore since it no longer comes through the signature.

plugins/plugin-core/src/types/index.ts
```diff
@@ -145,6 +145,54 @@ export interface StoragePluginHooks {
...
+       * @example "./keys/private-key.pem"
+       * @example "/secure/path/to/private-key.pem"
+       */
+      privateKeyPath: string;
+      /**
+       * Signature algorithm.
+       * Currently only RSA-SHA256 is supported.
+       * @default "RSA-SHA256"
+       */
+      algorithm?: "RSA-SHA256";
```
gronxb: If there’s only one algorithm, doesn’t it make the field unnecessary?

docs/content/docs/react-native-api/wrap.mdx
abdoolly: that makes sense as well.
I just felt it might be something someone wants to alert on or something.
but onError is enough and anyway it will still trigger to `onError`
will update.

plugins/plugin-core/src/types/index.ts
abdoolly: makes sense.
I will update

docs/content/docs/react-native-api/wrap.mdx
abdoolly: resolved.

plugins/plugin-core/src/types/index.ts
abdoolly: resolved.

abdoolly:
> It seems like I just need to check a few things! Most of it looks good — nice work.
> 
> By the way, is there any test evidence after applying the review? I’ll check it myself as well. 👀

Yes I did have a sample test app i test after changes.

packages/react-native/ios/HotUpdater/Internal/HotUpdaterImpl.swift
abdoolly: resolved.

abdoolly:
@gronxb all comments resolved.
I have also done testing to make sure unsigned and signed are working properly.
what do you think next steps to merge?

gronxb:
I’ll probably do a final review and merge it. I’ll do it as soon as I have time.

docs/content/docs/guides/bundle-signing.mdx
```diff
@@ -0,0 +1,409 @@
+---
+title: "Bundle Signing"
+description: "Cryptographically sign OTA bundles to verify integrity and prevent tampering"
+icon: Shield
```
gronxb, line 4:
```suggestion
icon: Shield
version: 0.23.0
```

tiagohs:
This looks amazing! ❤️  Is there backward compatibility for older APKs and IPAs that don't have the private key in the Manifest or Info.plist? I'd really like to use this, but a breaking change would be too costly for us, since the same bundle is deployed to both versions (with and without the key).

gronxb:
@tiagohs  As long as the bundle with the applied signing is properly configured with the target app version and fingerprint to prevent older versions of the app from downloading it, there should be no problem.
For already released versions, updates will proceed the same way as before — it’s just that they aren’t signed. Am I right? @abdoolly

tiagohs:
> As long as the bundle with the applied signing is properly configured with the target app version and fingerprint to prevent older versions of the app from downloading it, there should be no problem.

@gronxb My concern is exactly this. We already handle this with breaking changes, and it's costly to maintain separate bundles for old and new versions. This is particularly challenging when implementing performance improvements like `Brotli` compression, as `bundle.tar.br` files are not backwards compatible without the native client code being updated to decompress them.

In the long run, this turns into a costly issue to maintain, especially when dealing with a high volume of users (large installed base).

However, if contemplating/solving this is not feasible, that's Ok!

abdoolly:
> @tiagohs As long as the bundle with the applied signing is properly configured with the target app version and fingerprint to prevent older versions of the app from downloading it, there should be no problem. For already released versions, updates will proceed the same way as before — it’s just that they aren’t signed. Am I right? @abdoolly

To answer both your questions clearly.

 **Is there backward compatibility for older APKs and IPAs that don't have the public key in the Manifest or Info.plist?**

  Answer:
  No.

  Why:
  Old apps use HashUtils.verifyHash() which does a simple string comparison between the calculated SHA256 hash and the
  fileHash value.

  When a signed bundle is deployed, the fileHash is in the format sig:<base64_signature>. Old apps will:
  1. Calculate SHA256 of the bundle → abc123def... (hex string)
  2. Compare with sig:dG9rZW5fc2lnbmF0... (signature format)
  3. These don't match → HASH MISMATCH → Update rejected

------- 


 **Can the same bundle be deployed to both versions (with and without the public key)?**

  Answer:
  No.

  Why:

  | Bundle Type       | Old App (no public key)    | New App (with public key)       |
  |-------------------|----------------------------|---------------------------------|
  | Unsigned (<hash>) | ✅ Works - hash verified    | ❌ REJECTED - UnsignedNotAllowed |
  | Signed (sig:...)  | ❌ REJECTED - hash mismatch | ✅ Works - signature verified    |

  - Old apps treat fileHash as a plain hex hash. The sig: prefix causes a hash mismatch.
  - New apps with public key configured reject unsigned bundles (security rule: unsigned not allowed when signing enabled).

  There is no single bundle format that works for both app versions.

-----

 **Is this a costly breaking change similar to Brotli compression?**

  Answer:
  Yes.

  Why:
  The migration requires:
  1. Release new native app with public key embedded
  2. Wait for user adoption (old users cannot receive signed bundles)
  3. Either:
    - Maintain two deployment tracks (unsigned for old apps, signed for new apps)
    - Or accept that old app users receive no updates until they upgrade


There is a tradeoff which I will explain in the next comment.

abdoolly:
**Trade off Explanation and why this design choice** 

 Current Approach: Overloaded fileHash

  fileHash = "sig:<signature>"  (signed)
  fileHash = "<hex_hash>"       (unsigned)

  Pros:
  - No database migration required
  - No schema changes to any plugin (Supabase, Cloudflare D1, Firebase, etc.)
  - Simpler implementation

  Cons:
  - Breaking change - old apps fail on sig:... format (hash mismatch)
  - Cannot deploy same bundle to old and new apps

  ---
  Alternative Approach: Separate signature Column

  fileHash  = "<hex_hash>"        (always the hash)
  signature = "<base64>" | null   (new column)

  Pros:
  - Backward compatible - old apps ignore signature, verify fileHash 
  - Same bundle works for both old and new apps
  - Gradual migration possible

  Cons:
  - Requires database migration across ALL plugins
  - Schema changes to Supabase, Cloudflare D1, Firebase, PostgreSQL, etc.
  - More fields to maintain in types, APIs, and native code

abdoolly:
@tiagohs you can still easily maintain for signed and unsigned apps as @gronxb explained
without a very high cost if you made two separate github workflows for example.
with the below strategy.

**Dual Deployment Strategy**

Scenario: 
- Apps v3.0.0+ are signed
- Apps v2.x.x or lower remain unsigned.

GitHub Workflows


**Workflow 1: Unsigned (v2.x.x)**

```
  - name: Deploy unsigned bundle
    run: |
      npx hot-updater keys remove -y  // removes private keys from the app 
      npx hot-updater deploy -p ios -t 2.x.x --channel production


  Config: signing: { enabled: false }
```

  **Workflow 2: Signed (v3.0.0+)**

```
  - name: Deploy signed bundle
    run: |
      echo "${{ secrets.PRIVATE_KEY }}" > keys/private-key.pem
      npx hot-updater keys export-public -y // add private key to info.plist and strings.xml for signing
      npx hot-updater deploy -p ios -t 3.x.x --channel production


  Config: signing: { enabled: true, privateKeyPath: "./keys/private-key.pem" }
```


I think that could resolve your backward compatibility problem and just deploy the same update to both signed and unsigned apps easily just an additional bundle they could even happen in the same workflow without an additional workflow.

tiagohs:
Hmm, that's true. We're doing something similar to Brotli vs. Zip: different steps in the workflow generate two bundles, with two different targets.

All right, your explanation makes sense. There really wouldn't be any way to have backward compatibility with old binaries. When we migrated to `Expo Updates` with bundle signing, a breaking change was necessary then too.

Anyway, thanks for the detailed explanation! 😄

.changeset/witty-pillows-arrive.md
```diff
@@ -0,0 +1,10 @@
+---
+"@hot-updater/core": patch
+"hot-updater": patch
+"@hot-updater/react-native": patch
+"@hot-updater/cloudflare": patch
+"@hot-updater/firebase": patch
+"hot-updater-docs": patch
```
gronxb:
```suggestion
"@hot-updater/core": minor
"hot-updater": minor
"@hot-updater/react-native": minor
```

docs/content/docs/react-native-api/wrap.mdx
```diff
@@ -221,6 +221,30 @@ export default HotUpdater.wrap({
...
+## onError
+
+The `onError` callback handles all errors during the update process
+
+```tsx
+import {
+  HotUpdater,
+  getUpdateSource,
+  isSignatureVerificationError,
+  extractSignatureFailure,
```
gronxb:
```suggestion
  HotUpdater,
  getUpdateSource,
```

gronxb (approved):
Thank you. I manually ran all the tests. Everything went smoothly from start to finish. You’re amazing! 🚀🚀🚀

PR #723, opened 2025-12-02
feat(native)!!: notifyAppReady for auto rollback (invalid bundle)

gronxb:
## Context

When an incorrect bundle is deployed (such as a case where the app crashes immediately upon launch),
users may experience an infinite crash loop until they reinstall the app.
Therefore, we are implementing an automatic rollback system in the event of a crash.
Due to this change, there is a breaking change that requires `HotUpdater.wrap` to always be wrapped.

## 1. `getUpdateSource` removed

The `getUpdateSource` helper function has been removed.

```diff
- import { HotUpdater, getUpdateSource } from "@hot-updater/react-native";
+ import { HotUpdater } from "@hot-updater/react-native";

  export default HotUpdater.wrap({
-   source: getUpdateSource("https://api.com", { updateStrategy: "appVersion" })
+   baseURL: "https://api.com",
+   updateStrategy: "appVersion",
+   updateMode: "auto"
  })(App);
```

## 2. Manual Mode requires `HotUpdater.wrap`

Previously, it was possible to implement a custom update logic without using wrap, but now wrap is required even in those cases.

```diff
- export default App;
+ export default HotUpdater.wrap({
+   baseURL: "https://your-server.com",
+   updateMode: "manual"
+ })(App);
```

https://github.com/user-attachments/assets/5164e181-313d-4e1b-b469-16b9f742e4c8

### DIFF
--- a/packages/react-native/android/src/oldarch/HotUpdaterModule.kt
+++ b/packages/react-native/android/src/oldarch/HotUpdaterModule.kt
@@ -7,6 +7,7 @@ import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactApplicationContext
 import com.facebook.react.bridge.ReactMethod
 import com.facebook.react.bridge.ReadableMap
+import com.facebook.react.bridge.WritableNativeMap
 import com.facebook.react.modules.core.DeviceEventManagerModule
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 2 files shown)


gronxb:
@claude  please code review

gronxb:
@claude  please code review

PR #739, opened 2025-12-10
feat(iOS): Add tvOS support to HotUpdater

adamkoch:
All seems to work fine on tvOS, just needed to update the podspec.

### DIFF
--- a/.changeset/poor-brooms-remember.md
+++ b/.changeset/poor-brooms-remember.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+Add tvOS support to HotUpdater


gronxb:
Thanks for your contribution!

I have never worked with tvOS, and internally we also have to distribute builds separately for iOS and Android by platform.  
How does this work?  
If I distribute it for iOS, will it work?

adamkoch:
When using [react-native-tvos](https://github.com/react-native-tvos/react-native-tvos), anything that is "ios" will build for tvOS instead. It doesn't actually add a new "tvos" build target in RN. So from a hot-updater perspective, to release a tvOS patch it works fine just releasing an iOS patch as normal.

All this PR does is make the [iOS code files](https://github.com/gronxb/hot-updater/tree/main/packages/react-native/ios/HotUpdater) available to a tvOS build. Without the updated podspec, the tvOS build cannot see those files and the build will fail.

What this does NOT do is add the ability to target tvos specifically with a patch. So if you need to support both iOS and tvOS then a single "ios" patch would go out to both of them.

Note: there are some iOS-native APIs that are not supported on tvOS which will fail when building on tvOS but it built and ran fine for me so it doesn't look like you're using any of those.

gronxb:
Released in 0.24.3

adamkoch:
Fantastic, thanks!

PR #748, opened 2025-12-13
Use cachesDirectory on tvOS

adamkoch:
One more change is needed for tvOS (ref #739), not sure how it worked for me previously (maybe due to testing on simulator). On tvOS you aren't supposed to use or write to `documentDirectory`. Instead, it's recommended to use `cachesDirectory`.

This PR switches to use `cachesDirectory` on tvOS only so the bundle can be downloaded properly.


gronxb:
Released in 0.24.5

PR #749, opened 2025-12-13
fix(bare): add CocoaPods hermesc path for React Native 0.82+

nthtrung09it:
## Summary

React Native 0.82+ changed how Hermes is distributed:
- The `sdks/hermesc/` folder was removed from the react-native npm package
- hermesc is now distributed via CocoaPods at `ios/Pods/hermes-engine/destroot/bin/hermesc`

This PR adds support for the new CocoaPods path on macOS/iOS, fixing the "spawn node_modules/hermesvm/osx-bin/hermes ENOENT" error when using hot-updater with React Native 0.82+.

## Problem

When running `hot-updater deploy` with React Native 0.83, users get:

```
ExecaError: Command failed with ENOENT: node_modules/hermesvm/osx-bin/hermes --version
spawn node_modules/hermesvm/osx-bin/hermes ENOENT
```

The current `getHermesCommand()` function looks for hermesc at:
1. `react-native/sdks/hermesc/{os-bin}/hermesc` - ❌ Removed in RN 0.82+
2. `node_modules/hermes-engine/{os-bin}/hermesc` - ❌ Not installed by default
3. `node_modules/hermesvm/{os-bin}/hermes` - ❌ Fallback fails

## Solution

Added a check for the CocoaPods path (`ios/Pods/hermes-engine/destroot/bin/hermesc`) on macOS before falling back to the npm package paths.

## References

- [facebook/react-native@2e0bd13](https://github.com/facebook/react-native/commit/2e0bd13a2533fe7ab64125a95b9215b806018c6e) - RN 0.82: "Changed the source of hermesc binary to be an npm package"
- [facebook/react-native@27bb34c](https://github.com/facebook/react-native/commit/27bb34c006bb5daaac639d065aaa4c963528814e) - RN 0.83: "Use Hermes artifacts published independently from React Native"

## Test Plan

- [x] Tested with React Native 0.83.0 on macOS
- [x] Verified `hot-updater deploy` successfully finds hermesc at CocoaPods path
- [x] Confirmed Hermes compilation works (v0.14.0, HBC bytecode version 96)

### DIFF
--- a/plugins/bare/src/hermes.ts
+++ b/plugins/bare/src/hermes.ts
@@ -72,6 +72,7 @@ export async function getHermesCommand(cwd: string): Promise<string> {
   };
 
   // Since react-native 0.69, Hermes is bundled with it.
+  // Note: This path was removed in react-native 0.82+
   const bundledHermesEngine = path.join(
     getReactNativePackagePath(cwd),
     "sdks",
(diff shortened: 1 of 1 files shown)


gronxb:
Thanks for the contribution! However, I’m a bit concerned that this contribution might only work well on macOS for now.  
After checking a few things, it seems that pulling it directly from the hermes-compiler package would actually be the most accurate approach, so I’ll think a bit more about how to source it from there.

gronxb:
Based on your report, I changed the setup so that it fetches `hermes-compiler` instead of getting it from Pods. This should resolve the same issue. Thanks for letting me know! #754

gronxb:
Released in 0.24.6

PR #754, opened 2025-12-14
fix(bare): lookup hermesc path for React Native 0.82+

gronxb:
Related #749


gronxb:
@claude please code review

PR #780, opened 2025-12-30
Add support for hot-updater.config.mjs

pool611:
Duplacate hot-updater.config.cjs filename，the second `hot-updater.config.cjs` should be `hot-updater.config.mjs`


gronxb (approved):
Oh ! Thanks for your conribution

PR #795, opened 2026-01-14
feat(react-native): add brownfield support via HotUpdater.setReactHost()

gronxb:
Add support for brownfield apps that don't have ReactApplication:
- Add ReactHostHolder singleton to store ReactHost (newarch only)
- Add HotUpdater.setReactHost() extension function
- Add setJSBundle(bundleURL) and reload() overloads that use the stored ReactHost
- Update HotUpdaterImpl.reload() to try configured host before falling back to activity-based approach
- oldarch provides interface compatibility but returns false (not supported)


gronxb:
@claude please code review

PR #800, opened 2026-01-21
feat(cli): build:ios --simulator options

gronxb:
```
EXPERIMENTAL=1 pnpm hot-updater build:ios -s release --simulator
```

Extract `.app` file

### DIFF
--- a/packages/hot-updater/src/utils/native/createNativeBuild.ts
+++ b/packages/hot-updater/src/utils/native/createNativeBuild.ts
@@ -35,36 +35,4 @@ export const createNativeBuild = async ({
 
   const relativePath = path.relative(getCwd(), outputPath);
   p.log.info(`Artifact stored at ${colors.blueBright(relativePath)}`);
-
-  // If .app file exists, show its specific location
-  if (platform === "ios") {
-    const appFiles = await findAppFiles(outputPath);
-    if (appFiles.length > 0) {
-      p.log.info(`\n📱 .app files:`);
-      for (const appFile of appFiles) {
-        const relativeAppPath = path.relative(getCwd(), appFile);
-        p.log.info(`   ${colors.cyan(relativeAppPath)}`);
-      }
-    }
-  }
 };
-
-async function findAppFiles(dir: string): Promise<string[]> {
-  const appFiles: string[] = [];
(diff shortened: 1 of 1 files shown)


mym0404:
With all due respect, is not just putting `ios-simulator` in `destination` property of `NativeBuildSchema` for this feature? The initial concept of **scheme** must consists of the  `destination` property along with other properties such as build variant, configuration, and so on.

packages/apple-helper/src/utils/platform.ts
```diff
@@ -27,8 +27,8 @@ export const platformConfigs: Record<ApplePlatform, PlatformConfig> = {
     deviceSdk: "iphoneos",
     simulatorSdk: "iphonesimulator",
     supportsSimulator: true,
-    deviceDestination: "platform=iOS",
-    simulatorDestination: "platform=iOS Simulator",
+    deviceDestination: "generic/platform=iOS",
```
mym0404, line 30: We have to remain it as it was. `generic` is prepended while we resolve destinations later. [code](https://github.com/gronxb/hot-updater/blob/a39371592288951998eeac9e9e7d77464dff02df/packages/apple-helper/src/utils/destination.ts#L8). 

As a result, `generic` is added twice while it constructs `xcodebuild` params.

I came across it's somewhat complex to understand these processes.

packages/apple-helper/src/buildIos.ts
```diff
@@ -3,22 +3,55 @@ import type { NativeBuildIosScheme } from "@hot-updater/plugin-core";
...
 }): Promise<{ buildDirectory: string; buildArtifactPath: string }> => {
   await assertXcodebuildExist();
   const iosProjectRoot = path.join(getCwd(), "ios");
 
   const schemeConfig = await enrichNativeBuildIosScheme(_schemeConfig);
 
   const buildDirectory = await createRandomTmpDir();
+
+  // Simulator build: build .app directly without archiving
+  if (simulator) {
```
mym0404, line 27: You called `buildXcodeProject` to skip slow `xcodebuild` `archive` command. Is that what you intended? And, did you notice the practical build time difference?

mym0404:
@gronxb I appreciate you doing this job. If you don't mind, I might take this job on to handle the simulator destination separately during the `build:ios` process.

PR #805, opened 2026-01-31
fix(console): reference commit hash within git url hyperlink

ajchili:
When providing a `gitUrl` within the `console` configuration, it is improperly referenced within the Console UI. Rather than the commit hash being used, the function is used within the `href`.

Screenshot 2026-01-31 at 10 40 06 AM

----

Fixed this by calling the function when generating the `href` url.

https://github.com/user-attachments/assets/c05502aa-8ad2-4f39-a04e-f3b772179613

### DIFF
--- a/packages/console/hot-updater.config.ts
+++ b/packages/console/hot-updater.config.ts
@@ -46,4 +46,7 @@ export default {
       },
     ],
   }),
+  console: {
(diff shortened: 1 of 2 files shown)


gronxb (approved):
Thanks !!

PR #807, opened 2026-02-02
feat(ios): Add `bundle` parameter to `bundleURL()`

andrei-zgirvaci:
## Problem
Hot Updater's fallback bundle resolution hardcodes `Bundle.main`. This works for greenfield apps where the JS bundle is in the main app bundle, but fails for modern brownfield iOS apps where it's embedded in an XCFramework bundle.

## Solution
Add an optional `bundle` parameter to `bundleURL()` to specify which bundle contains the fallback JS bundle. This matches the [react-native-brownfield](https://oss.callstack.com/react-native-brownfield/docs/api-reference/react-native-brownfield/swift#properties) Swift API which already supports this pattern via its [`bundle`](https://github.com/callstack/react-native-brownfield/blob/main/packages/react-native-brownfield/ios/ReactNativeBrownfield.swift#L9) property.

## API
```swift
// Greenfield (unchanged)
HotUpdater.bundleURL()

// Modern brownfield
HotUpdater.bundleURL(bundle: xcframeworkBundle)
```

## Usage

```swift
// Initializes a bundle instance that points at the framework target
public let xcframeworkBundle = Bundle(for: InternalClassForBundle.self)
class InternalClassForBundle {}

// Configure Hot Updater with the previously created bundle instance
ReactNativeBrownfield.shared.bundleURLOverride = {
    return HotUpdater.bundleURL(bundle: xcframeworkBundle)
}
```

## Backward Compatibility
Fully backward compatible - defaults to `Bundle.main`.

## Related
https://oss.callstack.com/react-native-brownfield/docs/getting-started/ios#4-create-the-frameworks-public-interface
https://oss.callstack.com/react-native-brownfield/docs/getting-started/ios#7-initialize-react-native

### DIFF
--- a/.changeset/dirty-radios-roll.md
+++ b/.changeset/dirty-radios-roll.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+Add `bundle` parameter for XCFramework brownfield support on iOS


andrei-zgirvaci:
> Thanks 🫡

Thank you! When do you plan to release this btw?

gronxb:
Released in v0.25.8 !

andrei-zgirvaci:
> Released in v0.25.8 !

That's awesome! Thanks for the quick release!

PR #825, opened 2026-02-17
refactor(apple-helper): use npx pod-install for CocoaPods installation

mym0404:
## Summary
- Replace manual Gemfile/bundler detection and `pod install` / `bundle exec pod install` branching logic with `npx pod-install`
- Add 5-minute

### DIFF
--- a/packages/apple-helper/src/utils/cocoapods.ts
+++ b/packages/apple-helper/src/utils/cocoapods.ts
@@ -1,4 +1,4 @@
-import { getCwd, getReactNativeMetadatas, p } from "@hot-updater/cli-tools";
+import { getReactNativeMetadatas, p } from "@hot-updater/cli-tools";
 import { execa } from "execa";
 import fs from "fs";
 import path from "path";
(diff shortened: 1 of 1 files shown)


packages/apple-helper/src/utils/cocoapods.ts
```diff
@@ -14,54 +14,21 @@ export const installPodsIfNeeded = async (sourceDir: string): Promise<void> => {
...
-    if (!fs.existsSync(gemfilePath)) {
-      return false;
-    }
-
-    const gemfileContent = await fs.promises.readFile(gemfilePath, "utf-8");
-    return gemfileContent.includes("cocoapods");
-  } catch (_error) {
-    return false;
+    p.log.info("Installing CocoaPods started");
+    await execa("npx", ["-y", "pod-install", "--non-interactive"], {
```
gronxb, line 19: Considering the existing extra parameters, isn't the original one better? (Such as using prebuild or enabling the Hermes v1 flag)

Also, what about reducing npm dependencies?

mym0404: Thanks for your review. I see your point, but, at first, I keep passing the parameters using `env` argument. Moreover, I don't see it causes users to install extra dependencies on their project's `node_modules` because it's `npx`(did you point out different one?).

Even though I need to look up if `env` is passed down to `pod-install` package thoughfully.

I'll second-guess if it's really working and worth it.

mym0404: I'd like to consider adding options like Hermes v1 also.

gronxb: Since Hermes v1 becomes the default starting from 0.84.0, the flag will no longer be needed.  
From my perspective, flags appear and disappear all the time, so I don't think we need to expose them as our own options.  
I believe it's sufficient to pass extra arguments just like before.

gronxb: Even CocoaPods will probably move to SPM within the next two years. lol

https://github.com/facebook/react-native/blob/main/packages/react-native/Package.swift

mym0404: A big storm is coming for authors and developers. Unfortunately, I'm in both.

mym0404: Investigating further, I decided that this is useful. I believe there's something special about `pod-install` because I've been feeling like some official guides have indicated it's a way. [1](https://docs.expo.dev/bare/installing-expo-modules/#:~:text=%2D-,npx%20pod%2Dinstall,-%23%20Alternatively%2C%20the%20run)

On the plus side, it handles some kicky cases even such as automatic `pod update`. How are you seeing this?

gronxb: In my experience, running `RCT_USE_PREBUILT_RNCORE=1 RCT_USE_RN_DEP=1 bundle exec pod install` and running `RCT_USE_PREBUILT_RNCORE=1 RCT_USE_RN_DEP=1 npx pod-install` behaved differently.  
It felt like the environment variables were not being passed correctly, and that was my concern.  
If this works properly, then I'm fine with it.

gronxb: I'm relying on my memory, so it may not be accurate.

mym0404: It would be a good idea. I need double-check on it.

PR #835, opened 2026-02-25
fix(android): resolve resource lookup failure when applicationId differs from namespace

ksone02:
## Background

I run separate development and production environments, using hot-updater's channel feature to manage OTA updates independently for each. After upgrading to the latest version, I encountered a bug in `SignatureVerifier` where signature verification fails when downloading and verifying OTA bundle updates in the development build.

## Problem

`SignatureVerifier.getPublicKeyFromConfig()` uses `context.packageName` as the `defPackage` argument for `Resources.getIdentifier()`.

Since AGP 7.0, `namespace` (R class package) and `applicationId` (`context.packageName`) are officially separate concepts. When a project uses `applicationIdSuffix` for build variants (e.g. `.dev`, `.staging`), these two values diverge, causing `getIdentifier()` to fail with `resourceId == 0`.

When running the development variant, this throws `PublicKeyNotConfigured` even though `hot_updater_public_key` is correctly defined in `res/values/strings.xml`.

## Reproduction

1. Configure `productFlavors` with `applicationIdSuffix` in `build.gradle`:
```gradle
android {
    namespace "com.example"
      productFlavors {
        development {
          applicationIdSuffix ".dev" // context.packageName = "com.example.dev"
        }
     }
}
```
2. Add `hot_updater_public_key` to `res/values/strings.xml`
3. Enable signing in `hot-updater.config.ts`
4. Run the development variant
5. `PublicKeyNotConfigured` exception is thrown on every bundle update attempt

## Solution

Added a fallback that derives the namespace from `context.applicationInfo.className` when the initial lookup fails. The manifest shorthand `.MainApplication` is resolved against the namespace, so `applicationInfo.className` reliably contains the namespace as its package prefix.

```kotlin
if (resourceId == 0) {
    val appClassName = context.applicationInfo.className
    if (appClassName != null) {
        val namespace = appClassName.substringBeforeLast('.')
        if (namespace != context.packageName) {
            resourceId = context.resources.getIdentifier(
                "hot_updater_public_key",
                "string",
                namespace,
            )
        }
    }
}
```

Backward-compatible — the original lookup runs first, fallback only triggers on failure.

## References

- [Set the namespace - Android Developers](https://developer.android.com/build/configure-app-module#set-namespace)
- [Configure build variants - Android Developers](https://developer.android.com/build/build-variants)
- [Resources.getIdentifier() - Android Developers](https://developer.android.com/reference/android/content/res/Resources#getIdentifier)

### DIFF
--- a/packages/react-native/android/src/main/java/com/hotupdater/StringResourceUtils.kt
+++ b/packages/react-native/android/src/main/java/com/hotupdater/StringResourceUtils.kt
@@ -13,7 +13,10 @@ object StringResourceUtils {
      * @param resourceName The string resource name to resolve
      * @return The resolved resource ID, or 0 if not found
      */
-    fun getIdentifier(context: Context, resourceName: String): Int {
+    fun getIdentifier(
+        context: Context,
+        resourceName: String,
+    ): Int {
         var resourceId =
             context.resources.getIdentifier(
                 resourceName,


ksone02:
Good point. The same `context.packageName` pattern is used in other `getIdentifier` call sites, so they could be equally affected.

I can extract the fallback logic into a shared utility and apply it across all call sites. Let me know if you'd like me to update the PR accordingly.

gronxb:
Could you please update PR?

gronxb:
Released in v0.25.11

PR #850, opened 2026-03-02
feat: incremental PoC

gronxb:


gronxb:
To make this work, even if OTA bundles already exist, there’s a possibility that diffs might be generated based on the built-in bundle.  
Because of this, we either need to mix in the full bundle or maintain a separate `metadata.json` for each bundle, so it seems we need to approach it in a way that preserves the original state.

For example, in the initial change during a runtime channel switch, we use `minBundleId` even if OTA bundles are present.

https://github.com/gronxb/hot-updater/pull/858/changes/0cda15ce62f93b27081caede2448315b22881e8a

PR #852, opened 2026-03-02
fix(standalone): fetch channels from dedicated endpoint instead of paginated bundles

Hector-Zhuang:
I don’t think this is the correct implementation. For example, if the first 50 items are from the same channel, the other channels cannot be fetched.

### DIFF
--- a/.changeset/smart-buckets-change.md
+++ b/.changeset/smart-buckets-change.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/standalone": patch
+---
+
+fix(standalone): fetch channels from dedicated endpoint instead of paginated bundles


Hector-Zhuang:
Np, I'll add it later

Hector-Zhuang:
@gronxb The unit test has been added.

plugins/standalone/src/standaloneRepository.ts
```diff
@@ -133,8 +133,20 @@ export const standaloneRepository =
           };
         },
         async getChannels(): Promise<string[]> {
-          const result = await this.getBundles({ limit: 50, offset: 0 });
-          return [...new Set(result.data.map((b: Bundle) => b.channel))];
+          const { path, headers: routeHeaders } = routes.list!();
```
gronxb:
```suggestion
          const { path, headers: routeHeaders } = routes.list();
```

PR #857, opened 2026-03-08
fix(aws): use cache policies compatible with CloudFront plan billing

### LINKED ISSUE #855: CloudFront distribution config uses legacy settings incompatible with Plan billing mode
### Context
When attempting to switch the billing mode from `pay-as-you-go` to `Plan`, the following error is encountered:
> You're using configuration not available in this tier: legacy forwarded values, legacy minimum TTL, legacy maximum TTL

This prevents users from upgrading to the Plan tier entirely, as the CloudFront distribution deployed by Hot Updater appears to be provisioned with legacy cache behaviour settings that are not supported under the Plan billing model.

### Request
- Update the CloudFront distribution template to replace legacy forwarded values with a managed Cache Policy
- Remove inline legacy minimum TTL and legacy maximum TTL settings in favour of TTL values defined within the Cache Policy
- Ensure the updated configuration is compatible with the Plan billing tier so users can switch billing modes without manual intervention

gronxb:
Closed #855

### DIFF
--- a/plugins/aws/iac/cloudfrontDistributionConfig.ts
+++ b/plugins/aws/iac/cloudfrontDistributionConfig.ts
@@ -251,13 +251,10 @@ const buildCacheBehavior = (
 ): CacheBehavior => ({
   ...buildSharedBehavior(options.bucketName),
   PathPattern: template.pathPattern,
-  CachePolicyId: resolveCachePolicyId(
-    template.cachePolicy,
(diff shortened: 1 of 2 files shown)


gronxb:
@codex Please review the code to determine whether this PR is correct.

gronxb:
@codex review

PR #774, opened 2025-12-21
feat: Rollout feature with control from 1% to 100%

gronxb:
#2

## Summary
- replace the percentage/device-id rollout model with deterministic cohort rollout across core, React Native, server, console, and providers
- expose cohort-based manual targeting for QA/custom cohorts while keeping cached update routes aligned with provider cache keys
- add shared/provider/e2e coverage for the new rollout semantics and fold in CI follow-up fixes for iOS, Cloudflare, and example builds

## Breaking changes
- `rolloutPercentage` -> `rolloutCohortCount`
- `targetDeviceIds` -> `targetCohorts`
- update resolution input changes from `deviceId` to `cohort`
- remove `HotUpdater.setUserId()` / `HotUpdater.getUserId()`
- add `HotUpdater.setCohort()` / `HotUpdater.getCohort()`
- legacy header-based update checks now read `x-cohort` instead of `x-device-id`
- this is a breaking release with no compatibility alias layer or data migration for the old rollout fields

## Cohort model
- every device gets a deterministic numeric cohort string in the range `"1".."1000"`, derived from stable device identity
- the console still presents rollout as `0.0%..100.0%`, but persists `rolloutCohortCount: 0..1000`
- the first rollout for a bundle defines a stable bundle-id-seeded shuffle across `1..1000`; expanding or shrinking rollout reuses the same prefix
- `targetCohorts` supports numeric cohorts and custom lowercase/no-space cohorts, and always takes priority over gradual rollout
- custom cohorts do not participate in gradual rollout; they only receive updates via exact `targetCohorts` matches
- `setCohort("")` clears the manual override and falls back to the default numeric cohort

## What changed
- updated shared bundle types and update selection logic to use cohort fields and deterministic rollout selection
- replaced the native/public React Native user-id API with cohort APIs and propagated cohort through update-check requests
- updated server, JS fallback logic, provider handlers, and SQL procedures to use cohort suffix/header semantics
- aligned AWS, Cloudflare, Supabase, Firebase, and `@hot-updater/server` with the same eligibility behavior and shared suites
- replaced the console target-device-id UX with target-cohort UX and mapped slider values to `rolloutCohortCount`
- updated examples, migrations, docs, and generated interfaces to cohort terminology

### Bundle rollout fields
```ts
type Bundle = {
  rolloutCohortCount?: number | null;
  targetCohorts?: string[] | null;
};
```

### React Native
```ts
HotUpdater.setCohort("qa");
const currentCohort = HotUpdater.getCohort();
[...]

PR #860, opened 2026-03-10
fix: propagate shouldForceUpdate from intermediate OTA bundles

### LINKED ISSUE #572: Question: shouldForceUpdate flexibility and more
First of all thanks @gronxb for working on hot-updater, it's been really great using it so far!

So I have 2 questions about the update process:

1. Are there any plans for extending `shouldForceUpdate` to make it more flexible. If it was a string value we could have multiple ways to update (regular silent, pop-up for more important stuff, force restart for critical). I think JSON metadata for each bundle wouldn't hurt and could also replace `shouldForceUpdate` entirely
2. Let's imagine following bundles:
```
- bundleId3 // latest update, shouldForceUpdate = false;
- bundleId2 // fixes a critical crash, shouldForceUpdate = true;
- bundleId1 // current bundle, potentially crashes if it lets user into the app when `shouldForceUpdate === false`
 ```

Will the user be forced to update to bundle3? Unless I'm missing something that's a no, since `shouldForceUpdate` is set to `false` on the latest build that'll be fetched? I think you can see where I'm going with this, I know this adds some complexity but I'd love this to be handled properly, what do you think?

mindhells:
## Summary

Fixes #572

When a forced update (bundle 2) exists between the user's current bundle and a newer non-forced update (bundle 3), the `shouldForceUpdate` flag was silently lost — the server only returned the **latest** bundle's flag.

Now, if **any** enabled bundle between the user's current `bundleId` and the update candidate has `should_force_update = true`, the response's `shouldForceUpdate` is set to `true`. This ensures users cannot skip critical forced updates.

### Changes

Applied consistently across **all** server implementations:

- **`@hot-updater/server`** (`ormCore.ts`) — both `fingerprintStrategy` and `appVersionStrategy`
- **`@hot-updater/js`** (`getUpdateInfo.ts`) — in-memory implementation (also used by AWS Lambda)
- **`@hot-updater/firebase`** (`getUpdateInfo.ts`) — Firestore Cloud Function, uses an extra query for `should_force_update = true` intermediate bundles
- **`@hot-updater/cloudflare`** (`getUpdateInfo.ts`) — D1 SQL, added `intermediate_forced` CTE
- **`@hot-updater/postgres`** (SQL functions) — added `intermediate_forced` CTE to both PL/pgSQL functions

### How it works

For both the `currentBundle` path (user's bundle exists in candidates) and the `updateCandidate` path (user's bundle was disabled/removed):

1. After determining the update candidate, check if its `shouldForceUpdate` is already `true` → if so, no change needed
2. Otherwise, scan all enabled bundles with `id > bundleId` for any with `should_force_update = true`
3. If found, return the update candidate with `shouldForceUpdate` overridden to `true`
4. For `NIL_UUID` (first install), skip the intermediate check — there's no "skipping" scenario

### Tests

Added **10 new test cases** to the shared test suite (`setupGetUpdateInfoTestSuite.ts`), covering both `appVersion` and `fingerprint` strategies:

- Intermediate forced bundle propagates `shouldForceUpdate: true` (currentBundle path)
- All non-forced intermediates keeps `shouldForceUpdate: false`
- Latest bundle already forced — no change in behavior
- `NIL_UUID` first install — no intermediate check
- Current bundle not in candidates (updateCandidate path) — forced propagation

All 1768 tests pass across all implementations.

## Test plan

- [x] `pnpm -w build` — all 26 packages build successfully
- [x] `pnpm -w test` — 1768 passed, 1 skipped (pre-existing), 0 failed
- [x] `pnpm -w biome check` — no new errors (4 pre-existing in `examples/rock-repack`)

### DIFF
--- a/.changeset/propagate-intermediate-force-update.md
+++ b/.changeset/propagate-intermediate-force-update.md
@@ -1,0 +1,19 @@
+---
+"@hot-updater/server": patch
+"@hot-updater/js": patch
+"@hot-updater/firebase": patch
+"@hot-updater/cloudflare": patch
+"@hot-updater/postgres": patch
+---
+
+fix: propagate shouldForceUpdate from intermediate OTA bundles
+
+When a forced update (bundle 2) exists between the user's current bundle and
+a newer non-forced update (bundle 3), the forced flag was silently lost because
+only the latest bundle's `shouldForceUpdate` was returned.
+
+Now, if ANY enabled bundle between the user's current bundleId and the update
+candidate has `should_force_update = true`, the response's `shouldForceUpdate`
+is set to `true`. This ensures users cannot skip critical forced updates.
+
+Fixes #572
--- a/packages/server/src/db/ormCore.ts
+++ b/packages/server/src/db/ormCore.ts
@@ -188,12 +188,41 @@ export function createOrmDatabaseCore({
             latestCandidate &&
             latestCandidate.id.localeCompare(currentBundle.id) > 0
           ) {
+            // Propagate should_force_update if any intermediate bundle requires it
+            if (!latestCandidate.should_force_update) {
+              const hasIntermediateForced = sorted.some(
+                (b) =>
+                  b.id.localeCompare(bundleId) > 0 &&
+                  Boolean(b.should_force_update),
+              );
+              if (hasIntermediateForced) {
+                return toUpdateInfo(
+                  { ...latestCandidate, should_force_update: true },
+                  "UPDATE",
+                );
+              }
+            }
             return toUpdateInfo(latestCandidate, "UPDATE");
           }
           return null;
         }
 
         if (updateCandidate) {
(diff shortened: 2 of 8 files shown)


mindhells:
hey @gronxb let me know what you think about this PR, when you have the time 🙏

gronxb:
Thanks for the contribution. I understand the intention, but I think the console UX might need to change a bit if this feature is added, so I'm still thinking about it. The reason is that I want to avoid causing too much cognitive confusion. So I'd like to gather some reference use cases (like eas update). If this turns out to be cognitively easy to understand, I think I’ll be okay with it.

Secondly, I’m currently working on a large feature (#774) , so I don’t have time to review this right now. It looks like there will be significant changes in that PR, and the currently opened PRs may be affected. I’ll take a look at this as well after I finish my work. Thanks for your understanding.

PR #867, opened 2026-03-11
fix(android): use _jsBundleLoader backing field for Expo SDK 55+ compatibility

KimGnab:
## Problem

On Android with Expo SDK 55 (React Native 0.83), OTA updates via `forceUpdate` cause an
infinite reload loop. The app downloads the bundle successfully but fails to apply it,
falling back to the embedded bundle and re-triggering the update indefinitely.

**Logcat output:**
```
HotUpdater: Failed to setJSBundle with ReactHost: No field jsBundleLoader
  in class Lexpo/modules/ExpoReactHostFactory$ExpoReactHostDelegate;
```

## Root Cause

In Expo SDK 55, `ExpoReactHostDelegate.jsBundleLoader` was changed from a field-backed
property to a **Kotlin computed property** with a custom getter. This means there is no
Java field named `jsBundleLoader` — `getDeclaredField("jsBundleLoader")` throws
`NoSuchFieldException`.

The actual mutable backing field is `_jsBundleLoader`, which Expo intentionally maintains
for runtime bundle loader replacement:

```kotlin
// ExpoReactHostFactory.kt (Expo SDK 55)
// Keeps this `_jsBundleLoader` backing property for DevLauncher to replace its internal value
private var _jsBundleLoader: JSBundleLoader? = null
override val jsBundleLoader: JSBundleLoader
  get() {
    val backingJSBundleLoader = _jsBundleLoader
    if (backingJSBundleLoader != null) return backingJSBundleLoader
    // ... fallback logic
  }
```

**Source:** [`expo/packages/expo/android/.../ExpoReactHostFactory.kt#L60-L78`](https://github.com/expo/expo/blob/sdk-55/packages/expo/android/src/main/java/expo/modules/ExpoReactHostFactory.kt)

## Solution

Update `setJSBundleInternal()` to try `_jsBundleLoader` first (for Expo SDK 55+),
falling back to `jsBundleLoader` for backward compatibility with older Expo versions
and non-Expo `DefaultReactHostDelegate`.

This follows the same pattern used by Expo's own `expo-dev-launcher`, which accesses
`_jsBundleLoader` via reflection to replace the bundle loader at runtime:

```kotlin
// DevLauncherReactUtils.kt (Line 116-121)
if (reactHostDelegate.javaClass.canonicalName == EXPO_REACT_HOST_DELEGATE_CLASS) {
    reactHostDelegate.javaClass.setPrivateDeclaredFieldValue(
        "_jsBundleLoader", reactHostDelegate, newJsBundleLoader
    )
} else if (reactHostDelegate is DefaultReactHostDelegate) {
    DefaultReactHostDelegate::class.java.setPrivateDeclaredFieldValue(
        "jsBundleLoader", reactHostDelegate, newJsBundleLoader
    )
}
```

**Source:** [`expo/packages/expo-dev-launcher/android/.../DevLauncherReactUtils.kt#L96-L137`](https://github.com/expo/expo/blob/sdk-55/packages/expo-dev-launcher/android/src/debug/java/expo/modules/devlauncher/helpers/DevLauncherReactUtils.kt)

## Test Plan

- [x] Tested on physical Android device with Expo SDK 55 + hot-updater 0.26.0
- [x] Before fix: OTA `forceUpdate` triggered infinite reload loop (~1 reload/sec)
- [x] After fix: OTA bundle downloads, applies, and promotes to stable correctly
- [x] Backward compatibility: fallback to `jsBundleLoader` preserves support for
      `DefaultReactHostDelegate` and older Expo versions

### DIFF
--- a/.changeset/tender-cougars-visit.md
+++ b/.changeset/tender-cougars-visit.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+fix(android): use _jsBundleLoader backing field for Expo SDK 55+ compatibility


gronxb:
https://github.com/gronxb/hot-updater/releases/tag/v0.26.1

Issue #864, opened 2026-03-10
on android bundles are crashing more than 30% users


aravind-bazaar:
### Precheck

- [x] Infrastructure may have changed. Did you retry `hot-updater init`?
- [x] Is your Node version v20 or higher?
- [x] Are you aware that the update will only work properly if you `hot-updater deploy` after building the native app?

### Name, Version of hot-updater

0.25.11

### Version of react-native

0.81.5

### Platforms

Android

### Architecture

New Architecture(Fabric)

### What provider are you seeing the problem on? (multiple)

S3 & Lambda

### Expo App

- [ ] I am using Expo
- [ ] Project is managed with Expo [CNG](https://docs.expo.dev/workflow/continuous-native-generation/)(android, ios directories are included in `.gitignore`)

### Bundler

metro(hermes)

### What happened?

For android the budles are crashing for more than 30% users which makes it not worth to continue with hot updater

<img width="830" height="182" alt="Image" src="https://github.com/user-attachments/assets/e992f7a7-6d1b-4ea5-87b8-9f0ca06c58ac" />

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

gronxb:
https://github.com/gronxb/hot-updater/issues/453#issuecomment-3575365337

That's unfortunate. It seems that a crash occurs on Android when `reload` is called under certain conditions, so please refer to the related comment.

aravind-bazaar:
@gronxb wrt to comment mentioned no relevant crashes in firebase console. Mostly updates are applied in next launch using manual mode.

```tsx
async function checkForAppUpdate() {
    try {
      const updateInfo = await HotUpdater.checkForUpdate({ updateStrategy: 'appVersion' });

      if (updateInfo) {
        const repo = AdminRepositoryInstanceProvider.getClient();
        let response = { apply: true };

        try {
          response = await repo.applyCodepush(updateInfo.id, useUserStore.getState().cohort);
        } catch (error) {
          console.error('Failed to apply codepush:', error);
        }

        if (response.apply) {
          await updateInfo.updateBundle();
          analyticsInstance.logEvent(AnalyticEvent.CODEPUSH_DOWNLOAD, {
            bundle_id: updateInfo.id,
            file_hash: updateInfo.fileHash,
            file_url: updateInfo.fileUrl,
            message: updateInfo.message,
            should_force_update: updateInfo.shouldForceUpdate,
            status: updateInfo.status,
[...]

Issue #821, opened 2026-02-12
`await HotUpdater.reload()` causes the application to crash on Android


sawa-ko:
When I use this method, after downloading and applying the bundle, the app displays a gray screen and then reports that it has crashed.

I've seen this happen more on Android than on iOS, but I don't understand why this happens if it's supposed to just restart the app.

This is how my problem is replicated:

```ts
// hot-updater.config.ts

import { s3Storage } from '@hot-updater/aws';
import { expo } from '@hot-updater/expo';
import { withSentry } from '@hot-updater/sentry-plugin';
import { standaloneRepository } from '@hot-updater/standalone';
import { defineConfig } from 'hot-updater';

export default defineConfig({
  build: withSentry(
    expo({
      sourcemap: true,
    }),
    {
      org: '......',
      project: '.....',
      authToken: process.env.SENTRY_AUTH_TOKEN as string,
    }
  ),
  storage: s3Storage({
    region: process.env.AWS_REGION || 'us-east-1',
    bucketName: process.env.AWS_BUCKET_NAME,
  }),
  database: standaloneRepository({
    baseUrl: '.....',
  }),
  updateStrategy: 'appVersion',
  compressStrategy: 'tar.gz',
  signing: { enabled: true, privateKeyPath: './keys/private-key.pem' },
  console: {
    gitUrl: '.....t',
  },
});

```

```tsx
// _layout.tsx
const SentryLayout = Sentry.wrap(Layout) as ComponentType<object>;
export default HotUpdater.wrap({
  updateStrategy: 'appVersion',
  baseURL:
    process.env.EXPO_PUBLIC_UPDATER_URL,
  updateMode: 'auto',
})(SentryLayout);
```

> Using this hook

```tsx
import { useState } from 'react';

import { useResolveClassNames } from 'uniwind';

import { HotUpdater } from '@hot-updater/react-native';

export default function useUpdateNative() {
  const { toast } = useToast();
  const [isLoading, setIsLoading] = useState(false);

  const successNotificationIcon = useResolveClassNames('text-success mt-0.75');
  const dangerNotificationIcon = useResolveClassNames('text-danger mt-0.75');

  const checkUpdate = async () => {
    setIsLoading(true);

    try {
      const updateInfo = await HotUpdater.checkForUpdate({
        updateStrategy: 'appVersion'
      });

      if (!updateInfo) {
        setIsLoading(false);
        return null;
      }

      await updateInfo.updateBundle();

[...]

PR #869, opened 2026-03-13
feat(android): for safe reloading, Android reloads the process

gronxb:
**Related**

https://github.com/gronxb/hot-updater/issues/864
https://github.com/gronxb/hot-updater/issues/453
https://github.com/gronxb/hot-updater/issues/821

## Problem

This Android crash issue is not caused by HotUpdater itself, but rather by a conflict with another third-party library.

HotUpdater.reload() tears down the React instance and then recreates it.

However, if another third-party module tries to access the React instance during this process, a conflict can occur.

## Solution

To resolve this issue, the default reload strategy on Android is to restart the process.

Although this is heavier than simply reloading the React instance, it restarts from the splash screen, which is expected to effectively prevent such conflicts.

However, in the case of brownfield apps, reloading from an RN screen would completely restart the entire app. Therefore, the original mode is still provided.

You can use the previous behavior with:
```ts
HotUpdater. setReloadBehavior("reload");
```

### DIFF
--- a/packages/react-native/android/src/main/java/com/hotupdater/HotUpdaterImpl.kt
+++ b/packages/react-native/android/src/main/java/com/hotupdater/HotUpdaterImpl.kt
@@ -298,13 +298,9 @@ class HotUpdaterImpl {
      * @param reactContext The original context (preferably ReactApplicationContext to get current activity)
      */
     suspend fun reload(reactContext: Context) {
-        val reactIntegrationManager = ReactIntegrationManager(reactContext)
-
         try {
             withContext(Dispatchers.Main) {
-                val bundleURL = getJSBundleFile()
-                reactIntegrationManager.setJSBundle(bundleURL)
-                reactIntegrationManager.reload()
+                performReactReload(reactContext)
             }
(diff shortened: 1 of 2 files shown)


examples/v0.81.0/android/app/src/main/java/com/hotupdaterexample/MainApplication.kt
```diff
@@ -12,18 +12,13 @@ import com.facebook.react.defaults.DefaultReactNativeHost
...
 
 class MainApplication : Application(), ReactApplication {
-
   override val reactNativeHost: ReactNativeHost =
       object : DefaultReactNativeHost(this) {
         override fun getPackages(): List<ReactPackage> =
             PackageList(this).packages.apply {
-              // Packages that cannot be autolinked yet can be added manually here, for example:
-              // add(MyReactNativePackage())
+              add(ReloadCrashProbePackage())
```
gronxb, line 19: I successfully reproduced the issue using that package.

gronxb:
@codex review

gronxb:
@codex review

PR #874, opened 2026-03-15
fix(console): promote copy bundle id by original bundle id increment one

### LINKED ISSUE #865: Copy-promoted bundles cause infinite update loop due to hardcoded bundle ID in JS bundle
### Precheck

0.26.0

### Version of react-native

0.81.0

### Platforms

iOS, Android

### Architecture

New Architecture(Fabric)

Cloudflare (R2 + D1 + Workers)

### Expo App

- [ ] I am using Expo

### Bundler

metro

### What happened?

When a bundle is promoted to another channel via **copy** (not move), the new database entry gets a new `id` but shares the same `storageUri` — and thus the same physical JS bundle file — as the original.

The JS bundle file has the original bundle's ID hardcoded via `__HOT_UPDATER_BUNDLE_ID` (injected at build time). After the copied bundle is downloaded and applied, `getBundleId()` returns the **original** bundle's ID, not the copied one's:

https://github.com/gronxb/hot-updater/blob/6ea881a1a0a088ec7725cfd0f8b3b1c63235a6df/packages/react-native/src/native.ts#L215-L219

On the next update check, the client sends the original ID as its current `bundleId`. The server finds the copied bundle (with a newer ID) as an available update — and recognizes it as a new update again.

#### Impact

> **Critical (with `shouldForceUpdate`):** The app enters an **infinite reload cycle**. Since the bundle is already cached locally, the "download" completes instantly — but the app reloads immediately each time. If `fallbackComponent` is set, the user is **permanently stuck on the loading screen** with no way to use the app.

Without `shouldForceUpdate`, the same bundle is recognized as a new update on every update check. No visible breakage, but causes unnecessary update processing on every app launch.

#### Steps to Reproduce

1. Deploy bundle `A` (ID: `aaa-111`) to channel `production`
2. Copy-promote `A` to channel `staging` → creates bundle `B` (ID: `aaa-222`, same `storageUri`)
3. Client on `staging` applies bundle `B` (physically the same file as `A`, may resolve instantly from cache)
4. After reload, `getBundleId()` returns `aaa-111` (hardcoded in the JS file)
5. Client checks for update with `bundleId: "aaa-111"`
6. Server returns `B` as a new update (`aaa-222 > aaa-111`)
7. → Back to step 3 (infinite loop)

#### Affected Scope

All infrastructure backends are affected (Cloudflare D1, Supabase, AWS S3, Firebase, Standalone/PostgreSQL). The root cause is in the client-side `getBundleId()`, and all backends share the same `id >= bundleId` comparison logic for update checks.

Likely the same root cause as #861.

### Relevant log output

```shell

```

_No response_

### Code of Conduct

gronxb:
Closed #865 

The copy bundle regenerates a new bundle based on the timestamp of the original bundle ID. However, because of the random (rand) field, when comparing the generated bundle with the original bundle, it may be either larger or smaller with a 50% probability. Therefore, we use the same timestamp from the original bundle ID, but increase the rand field by 1.

### DIFF
--- a/.changeset/proud-buses-fail.md
+++ b/.changeset/proud-buses-fail.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/console": patch
+---
+
+fix(console): promote copy bundle id by original bundle id increment one


gronxb:
@codex review

PR #875, opened 2026-03-17
fix: apply UUIDv7 masking to copy-promoted bundle comparisons

### LINKED ISSUE #861: Bundle doesn't get marked as stable if promoted via copy
### Precheck

0.25.7

### Version of react-native

0.81.5

### Platforms

iOS, Android

### Architecture

New Architecture(Fabric)

Firebase

### Expo App

- [x] I am using Expo
- [x] Project is managed with Expo [CNG](https://docs.expo.dev/workflow/continuous-native-generation/)(android, ios directories are included in `.gitignore`)

### Bundler

metro

### What happened?

I'm currently doing the following workflow:
1. Push to `staging` channel
2. Promote build to `release` channel when all is good by copying (not moving)

The app on the `release` channel gets the update, but forcefully restarting the app (killing it from the task tray) and then opening the app again makes it go back to the previous version

If I directly push to the `release` channel, the update is persisted

--

I don't do `promote by moving` because those who has already received the update on `staging` channel gets stuck and a reinstall becomes needed

### Relevant log output

```shell

```

_No response_

### Code of Conduct

### LINKED ISSUE #865: Copy-promoted bundles cause infinite update loop due to hardcoded bundle ID in JS bundle

HealGaren:
## Summary

Fixes #865, fixes #861

Copy-promoted bundles share the same UUIDv7 timestamp but different random segments. This caused infinite update loops (#865) and native rollback after successful update (#861). This PR applies `maskUuidV7Rand()` masking across all bundleId comparisons — server, client, and native.

## Problem 1: infinite update loop (~50% chance)

Server-side `getUpdateInfo` compares the client's `bundleId` against DB records using strict equality to determine "already up-to-date". For copy-promoted bundles, the client's build-time `__HOT_UPDATER_BUNDLE_ID` differs from the DB's copy-promoted ID → server thinks it's a different version → returns an "update" for the same bundle → client downloads and installs again → loop.

## Problem 2: native rollback after update

When a copy-promoted bundle is installed:
1. `updateBundle()` stores the server-side bundleId as `stagingBundleId`
2. The JS bundle has a different `__HOT_UPDATER_BUNDLE_ID` baked in at build time (same timestamp, different random bits)
3. `notifyAppReady()` compares them with strict equality → **mismatch** → promotion fails
4. Next launch → native treats it as a crash → **rollback**

Confirmed via logcat:
```
notifyAppReady: bundleId mismatch (staging=019cf537-6412-74f2-af8f-01b0bce12522, current=019cf537-6412-777c-86ed-0e8e6c3665e1)
```

## How masking works

```
Input:  019cdd36-ab03-708b-a30a-d487efb35890
Lower:  019cdd36-ab03-7000-8000-000000000000  (maskUuidV7Rand)
Upper:  019cdd36-ab03-7fff-bfff-ffffffffffff  (maskUuidV7RandUpper)
```

Same timestamp → same masked range, regardless of random bits.

### Cross-source (server ID vs build-time ID) — masking applied

| Location | Original | Fix |
|----------|----------|-----|
| Android/iOS `notifyAppReady` | `stagingBundleId == currentBundleId` | `maskUuidV7Rand()` both sides |
| JS `native.ts` | `updateBundleId.localeCompare(getBundleId())` | `maskUuidV7Rand()` both sides |
| PostgreSQL / Supabase SQL | `b.id >= bundle_id`, `final_result.id != bundle_id` | See "SQL approach" below |
| JS plugin / ormCore / Cloudflare / Firebase | Various strict comparisons | `maskUuidV7Rand()` or `substr` |

### Same-source — no masking needed

| Location | Reason |
|----------|--------|
| Crash history (`BundleMetadata.kt/swift`) | Server ID → crash history → server ID check |
| Session state (`native.ts`) | Server ID within same session |

## SQL approach: range-based BETWEEN

PostgreSQL / Supabase stored procedures can't call TypeScript functions. The caller computes `masked_bundle_id_lower` and `masked_bundle_id_upper` and passes both:

```sql
-- Ordering (index-friendly, mask parameter only)
WHERE b.id >= masked_bundle_id_lower
WHERE b.id < masked_bundle_id_lower

-- Equality (BETWEEN = index range scan)
WHERE NOT (final_result.id BETWEEN masked_bundle_id_lower AND masked_bundle_id_upper)
WHERE b.id BETWEEN masked_bundle_id_lower AND masked_bundle_id_upper
```

## 🔵 Discussion: Supabase migration strategy

The SQL function signature changes from `bundle_id uuid` to `masked_bundle_id_lower uuid, masked_bundle_id_upper uuid`. This affects Supabase stored procedures and Edge Functions.

### Option A: Signature change (current implementation)

- Migration: `DROP FUNCTION` + `CREATE FUNCTION` with new params
- Edge Function also needs updating to pass masked params
- Could also create a v2 function name instead of dropping for backward compatibility
- Consistent with prior migrations in this project (0.13.0, 0.18.0, 0.21.0 all used DROP + CREATE)

### Option B: Body-only replacement

- Keep `bundle_id uuid` parameter unchanged
- Compute `masked_lower`/`masked_upper` inside SQL via `DECLARE` + `substr`/`||`
- Migration: `CREATE OR REPLACE` (no DROP)
- Edge Function unchanged
- Masking logic ends up in both SQL and TS, but the SQL side is just `substr` + concat

Open to either direction. Both require a migration; the difference is whether the Edge Function also changes.

## All backends modified

| Backend | File | Approach |
|---------|------|----------|
| Core | `packages/core/src/uuid.ts` | `maskUuidV7Rand` + `maskUuidV7RandUpper` |
| Android | `UuidUtils.kt` + `BundleFileStorageService.kt` | Masked `notifyAppReady` |
| iOS | `UuidUtils.swift` + `BundleFileStorageService.swift` | Masked `notifyAppReady` |
| JS client | `native.ts` | Masked `localeCompare` |
| JS plugin / ormCore | `getUpdateInfo.ts`, `checkForRollback.ts`, `ormCore.ts` | `maskUuidV7Rand()` in TS |
| Cloudflare D1 | `getUpdateInfo.ts` | `maskUuidV7Rand()` + `substr` in SQL |
| Firebase | `getUpdateInfo.ts` | `maskUuidV7Rand()` in TS |
| PostgreSQL | `plugins/postgres/sql/*.sql` + `getUpdateInfo.ts` | `BETWEEN lower/upper` range |
| Supabase | Migration + Edge Function | Pending discussion above |

## Test changes

- Test UUIDs changed from `00000000-0000-0000-0000-00000000000X` to `00000000-000X-0000-0000-000000000000` (timestamp-first format) — the old format placed the varying value in the random bits, which `maskUuidV7Rand` zeros out
- Added 4 copy-promoted bundle test cases (2 cases × app version / fingerprint strategies):
  - "returns null for same timestamp" — copy-promoted bundle (same timestamp, different random) should not trigger an update
  - "updates when newer version exists" — a genuinely newer bundle (different timestamp) should still be returned as an update, verifying that masking doesn't over-match

## Test plan
- [x] JS plugin: 152/152 passed
- [x] Cloudflare D1: 71/71 passed
- [x] Firebase: 98/98 passed
- [x] Server (elysia-libsql): 84/84 passed
- [x] PostgreSQL: 71/71 passed
- [x] Type check, Lint, Build: all clean
- [ ] Native (Android/iOS): manual testing needed

### DIFF
--- a/packages/react-native/android/src/main/java/com/hotupdater/BundleFileStorageService.kt
+++ b/packages/react-native/android/src/main/java/com/hotupdater/BundleFileStorageService.kt
@@ -343,7 +343,9 @@ class BundleFileStorageService(
         // Check for promotion (use masked comparison for copy-promoted bundles)
         if (isVerificationPending(metadata)) {
             val stagingBundleId = metadata.stagingBundleId
-            if (stagingBundleId != null && currentBundleId != null &&
+            if (
+                stagingBundleId != null &&
+                currentBundleId != null &&
                 UuidUtils.maskUuidV7Rand(stagingBundleId) == UuidUtils.maskUuidV7Rand(currentBundleId)
             ) {
                 Log.d(TAG, "App started successfully with staging bundle $currentBundleId, promoting to stable")
--- a/packages/react-native/android/src/main/java/com/hotupdater/UuidUtils.kt
+++ b/packages/react-native/android/src/main/java/com/hotupdater/UuidUtils.kt
@@ -25,11 +25,11 @@ object UuidUtils {
         // byte[7]      = rand_a (8 bits)
         // byte[8]      = variant (high 2 bits) | rand_b (low 6 bits)
         // bytes[9..15] = rand_b (56 bits)
-        bytes[6] = (bytes[6].toInt() and 0xf0).toByte() // keep version, clear rand_a high bits
-        bytes[7] = 0x00                                  // clear rand_a low bits
-        bytes[8] = (bytes[8].toInt() and 0xc0).toByte()  // keep variant, clear rand_b high bits
+        bytes[6] = (bytes[6].toInt() and 0xf0).toByte()
+        bytes[7] = 0x00
+        bytes[8] = (bytes[8].toInt() and 0xc0).toByte()
         for (i in 9 until 16) {
-            bytes[i] = 0x00                              // clear rand_b remaining bits
+            bytes[i] = 0x00
         }
 
         val out = bytes.joinToString("") { "%02x".format(it) }


gronxb:
First of all, I would like to apologize. Although I discussed this with you and you spent time helping me, I felt that I needed a more fundamental solution.

Thanks to you, I researched how not to include the bundleId in JavaScript, and as a result, I removed babel.config.js and came up with a way to ensure the bundleId properly on the native side.

Therefore, there is no longer a bundleId in JavaScript. With this approach, there will be no conflicts during promotion regardless of what bundleId is used.

Since all of these ideas came from discussions with you, I unfortunately have to close your PR, but I have credited you as a co-author of #893 as a token of my appreciation.

In fact, I would like to give you unlimited credit.

Thank you. This will be applied in version 0.29.0.

HealGaren:
No worries at all, and thank you for the kind words and the co-author credit on #893. I really appreciate the respect and thoughtfulness.

I'm glad our discussion helped lead to a cleaner direction. Removing the bundleId from JS entirely is a solid approach, and I imagine there are trade-offs to having the bundleId baked into the app, so it's great to see it resolved at a more fundamental level.

I'm actively using hot-updater in my product, and I'd love to keep contributing, whether that's reporting issues, joining discussions, or sending PRs when I can help. Looking forward to 0.29.0!

PR #878, opened 2026-03-19
feat(firebase): add immutable cache metadata and CDN URL support

mindhells:
## Summary
- Upload assets to Firebase/GCS with `Cache-Control: public, max-age=31536000, immutable` metadata, matching the behavior of the AWS S3 and Supabase storage plugins
- Add CDN URL support for Firebase managed deploys: when the `HOT_UPDATER_CDN_URL` environment variable is set on the Cloud Function, asset URLs are constructed from the CDN base URL instead of generating signed GCS URLs

## Test plan
- [x] Existing Firebase plugin tests pass (94/94)
- [x] Biome lint passes
- [ ] Verify uploaded objects have correct `Cache-Control` header in GCS console
- [ ] Verify CDN URL delivery by setting `HOT_UPDATER_CDN_URL` env var on Cloud Function and checking returned `fileUrl` in `/api/check-update` response
- [ ] Verify fallback to signed URLs when `HOT_UPDATER_CDN_URL` is not set


mindhells:
> Thanks for your contribution !
> 
> I have a few questions.
> 
> At what point do you expect HOT_UPDATER_CDN_URL to be injected?
> 
> Since this is a Firebase Function, should it be injected at the time of hot-updater init?
> 
> Also, where can I obtain (or provision) this CDN?

This would be a runtime env variable. As of now (the current state of this PR) is kind of a hidden feature (you'd need to update the firebase function manually). Passing it during `hot-updater init` sounds good, I can update the PR later if you think this is a good idea.

About the CDN: the whole point of having a CDN fronting the traffic that arrives into the firebase storage (instead delivering the signed URL directly to the users) is the cost. Egress is very expensive in gcloud: if you have a large user base this could mean thousands of $/month. The idea is that you can just configure yourself a CDN and use the firebase storage as backend, but then we need hot-update check-update API to serve the CDN URL. This is what we are doing now (we are wrapping the hot-updater firebase function impl. to do this), so I thought this is something that can be useful for other people.

Let me know if the idea of having support for CDN fronting is aligned with this project goal what would be the best approach add it, I'll be happy to update this PR.

gronxb:
For now, this approach is somewhat hidden, but it still seems possible to merge it. In the current self-hosted custom setup (https://hot-updater.dev/docs/custom/quick-start), you can plug in a plugin that resolves the storageUri, which works well when you are writing your own server. However, with managed Firebase, it’s unfortunate that modifying the code is difficult.
Although this is the best option for now, we will consider moving away from this structure in the future.

gronxb (approved):

gronxb (approved):

gronxb:
Thanks !

Issue #876, opened 2026-03-17
updateMode: 'manual' with next launch update silently crashing on android for 30% users


aravind-bazaar:
### Precheck

- [x] Infrastructure may have changed. Did you retry `hot-updater init`?
- [x] Is your Node version v20 or higher?
- [x] Are you aware that the update will only work properly if you `hot-updater deploy` after building the native app?

### Name, Version of hot-updater

0.25.11

### Version of react-native

0.81.5

### Platforms

Android

### Architecture

New Architecture(Fabric)

### What provider are you seeing the problem on? (multiple)

S3 & Lambda

### Expo App

- [ ] I am using Expo
- [ ] Project is managed with Expo [CNG](https://docs.expo.dev/workflow/continuous-native-generation/)(android, ios directories are included in `.gitignore`)

### Bundler

metro

### What happened?

I am using manual mode with next launch updates. The problem I am facing is codepush is getting downloaded but moved to crash_history with notifyAppReady event when bundle is recovered from crashed bundle. On Android the crash rate is  **_30%_**.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

gronxb:
Are you using reload when the force update flag is enabled? (If so, it will be resolved in version 0.27.0.)

Also, could you share the logic for checking and applying updates?

aravind-bazaar:
not using force update

```ts
import { CoachmarkOverlay, CoachmarkProvider } from '@edwardloopez/react-native-coachmark';
import { HotUpdater } from '@hot-updater/react-native';
import { AnalyticEvent, analyticsInstance } from '@modules/Analytics';
import { clearCart } from '@modules/Cart';
import { NavigationWrapper } from '@modules/Common';
import '@modules/Common/initUxCam';
import { CoachmarkStorage, isAndroid } from '@modules/Core';
import { useBottomNavStore } from '@modules/Home';
import { useUserStore } from '@modules/Login';
import { AdminRepositoryInstanceProvider } from '@modules/Network';
import { sendFcmTokenToBE } from '@modules/Notifications';
import { ErrorBoundary, Sentry } from '@modules/Sentry';
import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
import { useEffect, useMemo, useRef } from 'react';
import { AppState } from 'react-native';
import Config from 'react-native-config';
import { SafeAreaProvider } from 'react-native-safe-area-context';

interface AppProps {
[...]

PR #882, opened 2026-03-20
fix(react-native): improve rollback recovery

gronxb:
## Summary
- align Android and iOS rollback handling with Expo-style startup/error recovery boundaries
- make `notifyAppReady()` a read-only launch report API and remove public `PROMOTED` exposure
- add cross-platform crash monitoring, recovery markers, and launch-report/crash-history handling for rollback

https://github.com/user-attachments/assets/ed2fa6c9-807c-42a7-8d2d-1edc6635204c

https://github.com/user-attachments/assets/eb61d144-0783-4636-9c29-53e23b2e36c3

## As-Is
- Startup verification and rollback decisions are coupled to the JS-side `notifyAppReady(bundleId)` timing, so native recovery semantics depend on when JS reports readiness and the public `PROMOTED` state leaks an internal transition.
- Android and iOS do not share the same startup recovery boundary, which makes rollback behavior harder to reason about across JS callbacks, analytics, and docs.
- Fatal startup failures are not normalized into a single cross-platform launch report / crash-history flow, which weakens silent-crash recovery and retry safety.

## To-Be
- Native launch preparation finalizes bundle selection, pending rollback handling, and crash monitoring before JS reads launch state, aligned with Expo-style startup and error-recovery boundaries.
- `notifyAppReady()` becomes a read-only launch report API that returns only `STABLE` or `RECOVERED`, while native code promotes a staging bundle after first content appears.
- Android and iOS both persist launch reports and crash history, block previously crashed bundle IDs, and recover to the last stable bundle (or embedded bundle when needed) on the next launch.

### As-Is
1. A newly downloaded staging bundle launches first, but native only records that verification is pending and still depends on JS to acknowledge startup.
2. The startup contract is split across two `notifyAppReady` hops: JS must call `HotUpdater.notifyAppReady(bundleId)` to drive native promotion, and that lifecycle result may then be forwarded again through `resolver.notifyAppReady(...)` for external reporting.
3. If that JS-side `notifyAppReady` path is delayed or missed because the app crashes before mount, the wrapper is not wired, or a custom integration forgets the call, the bundle stays pending and rollback is inferred later from the missing callback on the next launch.

### To-Be
1. Before JS reads any launch state, native `prepareLaunch()` consumes the previous crash marker, applies rollback immediately when needed, selects the bundle to load, and starts crash monitoring for that launch.
2. Native crash/signal handlers stay armed through startup and continue watching for 10 seconds after the first render so delayed startup crashes are still captured.
3. On `CONTENT_APPEARED` / `RCTContentDidAppear`, native finalizes the launch transition itself: it marks the launch completed, clears `verificationPending`, flips native rollback state to `shouldRollback = false`, and treats the staging bundle as the verified current bundle without waiting for a JS acknowledgment.
4. JS `notifyAppReady()` becomes a read-only launch-report read (`STABLE` / `RECOVERED`) after native state has already been finalized, so reporting can no longer gate promotion or rollback semantics.

## User-Visible Behavior
- Previously, rollback was deferred until a later cold start noticed that `notifyAppReady(bundleId)` never arrived, so users had to hit the broken staging bundle first before recovery could happen. On Android that means at least one crashed launch before rollback; on iOS the old execution-counter flow could defer rollback even further.
- With this change, recovery is requested during the failing startup path instead of waiting for a manual relaunch: iOS triggers a recovery reload for JS startup failures, and Android schedules an automatic restart trampoline / watchdog-assisted relaunch after writing the crash marker.
- This is more accurate to describe as automatic recovery before the user manually reopens the app, not a universal same-process recovery. Android explicitly kills and restarts the process, and iOS signal/native crashes still fall back to consuming the crash marker on the next launch.

### DIFF
--- a/docs/content/docs/concepts/automatic-rollback.mdx
+++ b/docs/content/docs/concepts/automatic-rollback.mdx
@@ -1,53 +1,71 @@
 ---
 title: "Automatic Rollback"
-description: Hot Updater automatically detects crashed bundles and rolls back to stable versions.
+description: Hot Updater verifies newly installed bundles on first launch and automatically recovers to a working bundle if startup fails.
 icon: ShieldCheck
 version: "0.24.0"
 ---
 
 ## What is Automatic Rollback?
 
-Automatic rollback protects your app from getting stuck on crashing bundles. When a new update causes your app to crash, Hot Updater automatically detects the problem and rolls back to the last stable version.
+Automatic rollback is the safety mechanism that prevents a bad OTA bundle from trapping users in a crash loop.
 
-This happens automatically when you use `HotUpdater.wrap()`. No manual intervention needed.
+When a new bundle is installed, Hot Updater does not trust it immediately. It launches that bundle in a temporary state first, keeps the last known-good bundle as a fallback, and only treats the new bundle as safe after startup succeeds.
 
-## How It Works
+This happens automatically when you use `HotUpdater.wrap()`.
 
-When you deploy a new bundle, Hot Updater follows this process:
+## Core Mechanism
 
-1. **Download**: The new bundle downloads and is marked as "staging"
-2. **First Launch**: App restarts and loads the new bundle
-3. **Verification**:
-   - If the bundle loads successfully, `HotUpdater.wrap()` automatically verifies it
-   - The bundle is promoted to "stable" status
-4. **Rollback**: If the app crashes before verification, the next launch detects the crash and rolls back to the last stable bundle
+Hot Updater manages two roles during rollout:
 
-`HotUpdater.wrap()` handles verification automatically by calling `notifyAppReady()` when your app initializes. You don't need to call this function manually.
+- **Staging bundle**: the newly installed bundle waiting to be verified
+- **Stable bundle**: the last bundle that was already known to start successfully
 
-## Crash Detection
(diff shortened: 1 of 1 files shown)


gronxb:
E2E video attached via GitHub release asset.

- Video: https://github.com/gronxb/hot-updater/releases/download/pr-882-e2e-video-20260320-223242/pr882-ios-recovery.mp4
- Release page: https://github.com/gronxb/hot-updater/releases/tag/pr-882-e2e-video-20260320-223242
- Platform: iOS simulator (Release)
- Scenario: stable OTA (`019d0b58-44f8-77ee-ab84-3b68901064e5`) -> crash OTA (`019d0b6f-652d-7943-87d8-290438467a2a`) -> `Reload` -> automatic rollback / `RECOVERED`

gronxb:
@codex review

gronxb:
@codex review

gronxb:
@codex review

gronxb:
@codex review

gronxb:
@codex review

gronxb:
@codex review

gronxb:
@codex review

gronxb:
@codex review

gronxb:
Fixed #876

PR #900, opened 2026-04-03
fix(ios): improve archive validation and download persistence

### LINKED ISSUE #890: [iOS] Intermittent delay after update progress reaches 100% (possible decompression/storage bottlenecks)
### Precheck

0.25.7

### Version of react-native

0.81.5

### Platforms

iOS

### Architecture

New Architecture(Fabric)

S3 & Lambda

### Expo App

- [x] I am using Expo
- [x] Project is managed with Expo [CNG](https://docs.expo.dev/workflow/continuous-native-generation/)(android, ios directories are included in `.gitignore`)

### Bundler

metro(hermes)

### What happened?

On iOS, after the bundle download progress reaches 100%, we sometimes see an additional **3-5 second delay** before the app reloads. We do not see the same behavior on Android.

This is intermittent: even the same device may reproduce it only sometimes. It was first reported by internal TestFlight users while the app was already open in the foreground, so this does not appear to be related to screen lock / Data Protection.

After reviewing the iOS native implementation in `0.25.7`, I found three likely contributors in the native apply path.

---

#### 1. ZIP decompression on iOS appears to be fully memory-backed

**iOS** (`ZipDecompressionStrategy.swift`):
```swift
let zipData = try Data(contentsOf: URL(fileURLWithPath: file))
let zipEntries = try ZipContainer.open(container: zipData)
```

**Android** (`ZipDecompressionStrategy.kt`):
```kotlin
FileInputStream(filePath).use { fileInputStream ->
    BufferedInputStream(fileInputStream).use { bufferedInputStream ->
        ZipInputStream(bufferedInputStream).use { zipInputStream ->
```

On iOS, the archive is read into memory before extraction. Android, by contrast, uses a stream-based approach.

Because of that difference, iOS may be more sensitive to transient memory pressure during bundle application, which could explain why the delay is intermittent even on the same device.

Also, `TarGzDecompressionStrategy` / `TarBrDecompressionStrategy` appear to follow the same general pattern of loading compressed data into memory first, so switching compression format may not help with this specific symptom.

#### 2. `isValid()` appears to perform another full read before decompression

`ZipDecompressionStrategy.isValid()` also performs `Data(contentsOf:)` + `ZipContainer.open()` just to validate the archive format.

That means the same archive may be read and parsed once in `isValid()` and then again in `decompress()`.

From reading the code, it seems that lighter checks may already be enough at this stage:
- file exists
- minimum size
- ZIP magic bytes (`PK...`)

Then the actual structural validation could happen inside `decompress()` itself, where `ZipContainer.open()` is already called again.

#### 3. `URLSessionDownloadService` copies the downloaded archive before processing

`URLSessionDownloadService.swift` currently uses:
```swift
try FileManager.default.copyItem(at: location, to: destinationURL)
```

If `moveItem` is safe in this context, this may be avoidable extra I/O for large bundle files.

A `moveItem`-first approach with `copyItem` fallback may be worth considering, for example:
```swift
do {
    try FileManager.default.moveItem(at: location, to: destinationURL)
} catch {
    try FileManager.default.copyItem(at: location, to: destinationURL)
}
```

Android already seems to prefer rename/move first and only fall back to copy as a last resort in its bundle storage path.

---

### Why I think this matters

The delay happens **after progress reaches 100%**, so this does not look like a network issue.

From the iOS code path, it looks more likely that the extra time is being spent in one or more of these stages:
- archive validation
- decompression
- local file operations before the bundle is activated

### Suggested improvements

1. Consider switching iOS ZIP extraction to a stream/file-based implementation instead of `Data(contentsOf:)` + in-memory container parsing
2. Simplify `isValid()` so it only does lightweight format checks, and let `decompress()` handle the real structural validation
3. Consider trying `moveItem` first in `URLSessionDownloadService`, with `copyItem` as a fallback if needed
4. Optionally, review whether progress should reach `1.0` before all remaining post-download work is finished, since this makes the UI look "done" while native work is still ongoing

### Relevant log output

```shell
No error logs — this is a performance issue, not a crash or exception.
The visible delay happens between progress reaching 1.0 and the eventual reload.
```

_No response_

### Code of Conduct

gronxb:
## Summary
Fixes #890 on the iOS apply path.

This PR removes the main post-download bottlenecks that were keeping OTA apply work busy after progress had effectively reached 100%.

## Changes
- Replace the old memory-backed iOS ZIP path with a file-based `ZipArchiveExtractor` implemented directly in Swift.
- Move iOS `.tar.gz` and `.tar.br` extraction to a streaming flow via `StreamingTarArchiveExtractor`, writing a temporary `.tar` and extracting entries incrementally.
- Keep ZIP and TAR.GZ validation cheap and signature-based, and only attempt TAR.BR as the final extraction fallback.
- Use move-first archive persistence in `URLSessionDownloadService`, with copy fallback only when `moveItem` is not possible.
- Re-map apply progress so download, verification, extraction, bundle validation, and activation are reported in separate ranges instead of jumping to `1.0` before native work is done.
- Defer old bundle cleanup off the visible success path.
- Harden the native extractors against path traversal, symbolic link entries, ZIP64 inputs, checksum mismatches, and corrupted TAR archives.
- Add Swift archive extraction tests and committed fixtures for `zip`, `tar.gz`, and `tar.br`, including:
  - real deploy-produced fixtures
  - large extracted-payload fixtures
  - archive-sized 300 MiB fixtures stored as split parts for git-safe reuse
- Add `pnpm test:swift` at the repo root and run it in the iOS integration workflow.

## Why this fixes #890
- The iOS apply path no longer depends on full-archive `Data(contentsOf:)` reads for ZIP / TAR extraction.
- ZIP validation no longer performs a second full archive read before extraction.
- Downloaded archives are persisted with move-first semantics instead of always copying.
- Heavy native work after download completion is reduced to file-based extraction plus final activation steps, instead of full compressed-payload loads and synchronous cleanup on the success path.
- Reported progress now matches actual completion of verification, extraction, and activation work.

## Validation
- `pnpm test:swift`
  - `pnpm -w build`
  - `pnpm -w test:type`
  - `pnpm -w lint`
  - `pnpm -w test`
  - `pnpm -w test:integration`

### DIFF
--- a/e2e/maestro/flows/archive-300mb-ota.yaml
+++ b/e2e/maestro/flows/archive-300mb-ota.yaml
@@ -79,6 +79,10 @@ onFlowComplete:
     env:
       APP_ID: ${APP_ID}
 - runFlow:
+    file: subflows/navigate-top.yaml
+    env:
+      APP_ID: ${APP_ID}
+- runFlow:
     file: subflows/assert-text-by-id.yaml
     env:
       APP_ID: ${APP_ID}
--- a/examples/v0.81.0/App.tsx
+++ b/examples/v0.81.0/App.tsx
@@ -31,7 +31,9 @@ const notify = proxy<{
 
 const DEFAULT_APP_BASE_URL = "http://localhost:3007/hot-updater";
 const HOT_UPDATER_BASE_URL = HOT_UPDATER_APP_BASE_URL || DEFAULT_APP_BASE_URL;
-const E2E_SCENARIO_MARKER = "builtin-ios-maestro";
+const E2E_SCENARIO_MARKER = "runtime-channel-beta-maestro";
+const E2E_LARGE_ARCHIVE_ASSET_MANIFEST_PATH =
+  "assets/src/test/_fixture-archive-300mb-random.bmp";
 
 function maybeCrashForE2E() {
   /* E2E_CRASH_GUARD_START */
@@ -39,6 +41,13 @@ function maybeCrashForE2E() {
 }
 
 maybeCrashForE2E();
+
+function loadE2EDeployBundleAssets() {
+  /* E2E_DEPLOY_ASSET_GUARD_START */
+  /* E2E_DEPLOY_ASSET_GUARD_END */
+}
+
+loadE2EDeployBundleAssets();
 
 const getGlobalBaseUrl = (): string | null => {
   const maybeFn = Reflect.get(globalThis, "HotUpdaterGetBaseURL");
@@ -220,6 +229,9 @@ function App(): React.JSX.Element {
   }, []);
 
   const manifestAssetEntries = Object.entries(runtimeSnapshot.manifest.assets);
+  const hasLargeE2EAsset = manifestAssetEntries.some(
+    ([fileName]) => fileName === E2E_LARGE_ARCHIVE_ASSET_MANIFEST_PATH,
+  );
   const statusPayload = JSON.stringify(notifyState, null, 2);
   const launchStatusText = `Current Launch Status: ${notifyState.status ?? "null"}`;
   const crashedBundleText = `Current Crashed Bundle ID: ${notifyState.crashedBundleId ?? "null"}`;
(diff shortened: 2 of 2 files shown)


gronxb:
@codex review

gronxb:
@codex review

PR #904, opened 2026-04-04
fix(standalone): stale data due to module-level singleton configPromise and shared changedMap across requests

### LINKED ISSUE #903: [console] Stale data due to module-level singleton configPromise and shared changedMap   across requests
This is what claude helped me troubleshoot, I often encounter that after deleting a bundle in the console, the list doesn't refresh, or after refreshing it still has the old data

## Environment

| Package | Version |
|---|---|
| `@hot-updater/console` | `0.29.2` |
| `@hot-updater/standalone` | `0.29.2` |
| `@hot-updater/plugin-core` | `0.29.1` |
| Database | `standaloneRepository` (HTTP mode) |

---

### Root Cause

In `config.server.mjs` (console package), `databasePlugin` is created once as a module-level singleton and never reset:

```ts
var configPromise = null;

var prepareConfig = async () => {
  if (!configPromise) configPromise = (async () => {
    const databasePlugin = await config.database(); // created once, never reset
    // ...
  })();
  return configPromise;
};
```

Inside `createDatabasePlugin` (plugin-core), each instance owns a `changedMap` that is shared across **all** incoming server requests:

```ts
const changedMap = new Map(); // never reset unless commitBundle succeeds

async updateBundle(targetBundleId, newBundle) {
  const pendingChange = changedMap.get(targetBundleId);
  if (pendingChange) {
    // uses stale in-memory data — no HTTP request to fetch current state
    changedMap.set(targetBundleId, {
      operation: pendingChange.operation,
      data: mergeBundleUpdate(pendingChange.data, newBundle),
    });
    return;
  }
  // only reaches here if changedMap has no entry for this id
  const currentBundle = await getMethods().getBundleById(targetBundleId);
  markChanged("update", mergeBundleUpdate(currentBundle, newBundle));
}
```

`changedMap.clear()` is only called inside `commitBundle` after a successful commit. If `commitBundle` fails (network error, server 5xx, etc.), the map is permanently dirty for the lifetime of the process.

### Failure Scenario

1. Request A calls `updateBundle(id, patch)` → writes to `changedMap`
2. `commitBundle()` throws → `changedMap` is **not cleared**
3. Request B (different tab / user) calls `updateBundle(id, patch2)` → hits the `pendingChange` branch, merges against stale data from Request A **without ever fetching current state from the API**
4. `commitBundle()` sends incorrect merged data to the server

> **Current workaround:** restart the console process to clear the singleton.

### Suggested Fix

**Option A** — reset `configPromise` on init error so the next request can reinitialize:

```ts
if (!configPromise) configPromise = (async () => {
  try {
    // ...
  } catch (e) {
    configPromise = null; // allow re-initialization on next request
    throw e;
  }
})();
```

**Option B (preferred)** — create a fresh `databasePlugin` per request to avoid a shared `changedMap`:

```ts
var prepareConfig = async () => {
  const config = await getConfig(); // config itself can still be cached
  const databasePlugin = await config.database(); // fresh instance per request
  // ...
};
```

Option B is safer for concurrent requests since each request gets its own isolated `changedMap`.

---

### Root Cause

`getBundles()` does not forward `limit`/`offset` query params to the API:

```ts
async getBundles(options) {
  const { where, limit, offset = 0, orderBy } = options ?? {};
  const response = await fetch(buildUrl("/api/bundles")); // no limit/offset in URL
  const filteredBundles = (await response.json())
    .filter(bundle => bundleMatchesQueryWhere(bundle, where)); // client-side filter on truncated data
  // ...
}
```

If the server defaults to `limit=50`, only the first 50 bundles are ever returned. Client-side filtering on this truncated set causes bundles beyond position 50 to appear nonexistent in the console.

### Suggested Fix

Forward pagination params to the API:

```ts
async getBundles(options) {
  const { where, limit, offset = 0, orderBy } = options ?? {};
  const { path, headers: routeHeaders } = routes.list();
  const url = new URL(buildUrl(path));
  if (limit) url.searchParams.set("limit", String(limit));
  url.searchParams.set("offset", String(offset));
  const response = await fetch(url.toString(), {
    method: "GET",
    headers: getHeaders(routeHeaders),
  });
  // ...
}
```

gronxb:
Fixed #903

### DIFF
--- a/docs/content/docs/database-plugins/standalone.mdx
+++ b/docs/content/docs/database-plugins/standalone.mdx
@@ -193,8 +193,8 @@ interface Bundle {
 
 ### List Bundles Contract
 
-`GET /api/bundles` must accept pagination and filter query params and return a
-paginated JSON body.
+`GET /api/bundles` must accept the pagination and filter query params used by
+the console and return a paginated JSON body.
 
 Supported query params:
 
@@ -202,19 +202,6 @@ Supported query params:
 |-------------|------|-------|
 | `channel` | `string` | Exact match |
 | `platform` | `"ios" \| "android"` | Exact match |
-| `enabled` | `"true" \| "false"` | Boolean filter |
-| `idEq` | `string` | Exact ID match |
-| `idGt` | `string` | ID greater than |
-| `idGte` | `string` | ID greater than or equal |
-| `idLt` | `string` | ID less than |
-| `idLte` | `string` | ID less than or equal |
-| `idIn` | `string[]` | Repeat the query param for multiple IDs |
-| `targetAppVersion` | `string` | Exact match |
-| `targetAppVersionNull` | `"true"` | Filter rows where `targetAppVersion` is `null` |
-| `targetAppVersionIn` | `string[]` | Repeat the query param for multiple values |
-| `targetAppVersionNotNull` | `"true"` | Filter rows where `targetAppVersion` is not `null` |
-| `fingerprintHash` | `string` | Exact match |
-| `fingerprintHashNull` | `"true"` | Filter rows where `fingerprintHash` is `null` |
 | `limit` | `number` | Page size |
 | `offset` | `number` | Page offset |
 
--- a/packages/server/src/handler.spec.ts
+++ b/packages/server/src/handler.spec.ts
@@ -207,47 +207,6 @@ describe("createHandler", () => {
     );
   });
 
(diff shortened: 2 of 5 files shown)


gronxb:
@codex review

gronxb:
@codex review

gronxb:
@codex review

gronxb:
@codex review

PR #911, opened 2026-04-05
feat: Introduce bundle diffing (~500kb OTA Update) 

gronxb:
https://github.com/user-attachments/assets/22b90f58-91ff-40e1-8a6e-ab2fbfa5a2a6

## What changed
- add signed manifest-based diff update support across deploy, server, and the React Native runtime
- upload `manifest.json` plus per-file assets during deploy, while keeping the archive artifact as a fallback path
- add native install flows on iOS and Android that verify the signed manifest, reuse unchanged files, download only changed assets, and fall back to the archive when verification or reuse fails
- add regression coverage for diff payload generation and native spec handling
- add a changeset for the public packages involved

## Why
`manifest.json` is now the signed source of truth for changed-files-only OTA installs. This makes the first diffing phase usable before introducing `bsdiff`, while still preserving the existing archive path for older bundles and failure recovery.

This also fixes a real deploy bug found during E2E verification: Hermes outputs such as `index.ios.bundle.hbc` were uploaded with the local basename, while diff manifests referenced `index.ios.bundle`, which caused asset URLs to 404.

## Impact
- clients can install smaller OTA updates by reusing unchanged files from the current bundle directory
- signed manifest verification gates the diff path, so tampered manifests or asset mismatches fall back to the archive flow
- older deployments without manifest metadata continue to work through the existing archive-only behavior

## Validation
- `pnpm -w lint`
- `pnpm -w test -- packages/hot-updater/src/commands/deploy.spec.ts packages/server/src/runtime.spec.ts packages/react-native/src/native.spec.ts`
- `pnpm -w build`
- end-to-end verification against the running hono-s3 update server with `examples/v0.81.0`
  - iOS stable diff install verified with bundle `019d5bc3-0406-7ee4-85f9-a318a03dcf5e`
  - iOS crash rollback verified with crashed bundle `019d5bc8-1a97-7a9b-9182-ab83db4bc1c7`
  - Android stable diff install verified with bundle `019d5bfc-f023-7b7f-b1ac-ccfed2a5923a`
  - Android crash rollback verified with crashed bundle `019d5c00-b73c-7534-a59f-8e6974ae8685`


gronxb:
@hot-updater-e2e

nthtrung09it:
@gronxb  can't wait to see it merged :)

dusty-cai:
@gronxb: Let me know if you'd like eyes on this, I know going solo on a project can be rough

gronxb:
@dusty-cai Thanks for the encouragement! As you said, it took a while since I was working alone, but it's really in the final stage now, and this feature will be released this week.

PR #916, opened 2026-04-07
fix(android): oldarch sync methods crash with WritableNativeMap/Array return types

anncwb:
## Summary

On React Native 0.72 (old architecture, `newArchEnabled=false`), the app crashes on startup with:

```
Exception in HostObject::get for prop 'HotUpdater': java.lang.RuntimeException: Got unknown return class: WritableNativeMap
Exception in HostObject::get for prop 'HotUpdater': java.lang.RuntimeException: Got unknown return class: WritableNativeArray
```

### Root Cause

`JavaMethodWrapper.returnTypeToChar()` in React Native's bridge uses **exact class matching**:

```java
if (returnClass == WritableMap.class) return 'M';     // interface ✅
else if (returnClass == WritableArray.class) return 'A'; // interface ✅
else throw new RuntimeException("Got unknown return class: " + returnClass.getSimpleName());
```

The oldarch `HotUpdaterModule.kt` declares **concrete classes** as return types:

```kotlin
override fun notifyAppReady(): WritableNativeMap   // ❌ WritableNativeMap != WritableMap
override fun getCrashHistory(): WritableNativeArray // ❌ WritableNativeArray != WritableArray
override fun getManifest(): WritableNativeMap       // ❌ WritableNativeMap != WritableMap
```

Since `WritableNativeMap.class != WritableMap.class`, the bridge throws `RuntimeException` during **module registration** (not method invocation), preventing the entire `HotUpdater` module from loading.

### Fix

Changed all three sync methods to return JSON strings. The JS side already handles this with existing string fallbacks:

```typescript
// native.ts - all three methods already have string fallback
if (typeof result === "string") { return JSON.parse(result); }
```

### Environment

- `react-native`: 0.72.17
- `@hot-updater/react-native`: 0.29.3
- Architecture: Old (bridge), `newArchEnabled=false`
- JS Engine: Hermes
- Platform: Android

### DIFF
--- a/.changeset/thick-terms-guess.md
+++ b/.changeset/thick-terms-guess.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+fix(android): oldarch sync methods crash with WritableNativeMap/Array return types


gronxb:
Thanks ! Could you run ktlint ?

anncwb:
> Could you check ktlint ?

done

PR #930, opened 2026-04-09
feat: add stacked HBC bsdiff patches

gronxb:
## What changed
- add HBC patch metadata and delivery wiring so the server can attach a `bsdiff` patch descriptor to changed Hermes bundle assets when the current bundle matches the selected diff base
- add console support to pick a base bundle, generate and upload an HBC diff artifact, and visualize bundle relationships as a stacked chain in the bundles table and metadata view
- add native `bspatch` application support on iOS and Android, including bridge parsing for the patch descriptor and runtime fallback to the normal asset download path
- add tests for console diff creation, native patch payload parsing, and server patch descriptor responses
- add a changeset for the affected package release

## Why
PR #911 adds signed manifest-based diff installs by reusing unchanged files. This follow-up stack adds the next step for Hermes bytecode bundles: ship a binary patch against a chosen base bundle instead of re-downloading the full HBC file.

That keeps the scope focused on explicit diff-base selection and native patch application. Backfill generation for older bases is intentionally deferred to a later PR.

## Impact
- console operators can create a diff artifact tied to a specific base bundle and see that relationship in the UI
- compatible clients can apply a `bsdiff` patch to the Hermes bundle natively on iOS and Android
- clients that are not on the expected base still fall back to the existing manifest/archive flows

## Validation
  - `build`
  - `test:type`
  - `lint`
  - `test`
  - `test:integration`

## Stack
- base PR: #911
- this PR targets `feature/diffing`
- backfill logic is out of scope for this PR


muhamedzeema:
Hey @gronxb, I’m really excited to see this PR get merged. let me know if you need any help!

gronxb:
@hot-updater-e2e standalone

PR #940, opened 2026-04-13
fix(server): keep target cohorts additive to rollout

gronxb:
## Summary
- make `Target Cohorts` additive to numeric rollout instead of overriding it
- add shared/provider coverage and a Maestro scenario for `0% + qa`, `50% + qa`, included numeric cohorts, and excluded numeric cohorts
- update the docs and console copy to explain the additive rollout model
- fix Android restart behavior so a manual cohort override such as `qa` survives install/reload flows
- add a new Supabase migration file for the eligibility SQL update

## Why
`Target Cohorts` is most useful when QA or tester cohorts should keep receiving the same bundle while gradual numeric rollout continues for everyone else. This PR makes that behavior explicit in the implementation, coverage, and docs.

## Validation
- `pnpm -w e2e:maestro`

## Notes
- iOS passed the new scenario as-is
- Android needed the native persistence fix because the cohort override could be lost across restart when the app relaunched immediately after OTA install
- the Supabase change is shipped as a new migration: `plugins/supabase/supabase/migrations/20260414000000_hot-updater_0.30.0.sql`


gronxb:
Updated this PR with the additive `Target Cohorts` follow-up.

Included in commit `547fa4821`:
- persist Android cohort overrides synchronously across restart so `qa` survives install/reload flows
- add Android bundle/baseURL/manifest resolution from the current launch selection with unit coverage
- add the `target-cohorts-with-rollout` Maestro scenario for `0% + qa`, `50% + qa`, included numeric cohorts, and excluded numeric cohorts
- refresh the rollout docs and console copy to explain additive rollout semantics
- add a new Supabase migration file: `20260414000000_hot-updater_0.29.8.sql`

Validation completed locally:
- `pnpm -w e2e:maestro`

PR #945, opened 2026-04-17
fix(native): launched bundle identity reporting

gronxb:
## What changed

- Treat missing native launch selection as the built-in bundle on Android and iOS instead of falling back to persisted bundle metadata.
- Keep runtime identity owned by native: `getBundleId()`, `getManifest()`, and `getBaseURL()` only return downloaded bundle data when HotUpdater actually prepared that bundle for launch.
- Change the native module boundary for `getBaseURL()` to return `null` when unavailable instead of using an empty-string sentinel. Storage internals still use empty strings to keep this change narrow.
- Add unit coverage for native-entrypoint-missing and pending-staging cases across JS, Android, and iOS storage.
- Ignore local `.omx/` state in git.

## Why

Apps can download an OTA bundle even when `AppDelegate.swift` or `MainApplication.kt` is missing the HotUpdater native bundle setup. In that state the app still runs the embedded bundle, but persisted staging metadata could make `getBundleId()` report the newly downloaded bundle. That made runtime identity inconsistent with the code actually executing.

This fix keeps that decision in native code rather than warming JS getter caches before downloads. It also makes the native `getBaseURL()` absence value match the public JS API (`null`).

## Validation

- `pnpm --filter @hot-updater/react-native test -- src/native.spec.ts --run`
- `./gradlew testDebugUnitTest --tests com.hotupdater.BundleFileStorageServiceTest --tests com.hotupdater.HotUpdaterImplTest`
- `swift test --package-path packages/react-native/ios/HotUpdater`
- `pnpm --filter @hot-updater/react-native test:type`
- `pnpm -w lint`

## Not tested

- Device E2E with intentionally missing native setup.

### DIFF
--- a/.changeset/loud-seahorses-arrive.md
+++ b/.changeset/loud-seahorses-arrive.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+fix(native): launched bundle identity reporting
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,4 @@ plugins/supabase/runtime-acceptance-*/
 .codex/
 
 .env.hotupdater
+.omx/
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -67,7 +67,7 @@
     "tslib": "^2.8.1",
     "tw-animate-css": "^1.4.0",
     "typescript": "^6.0.2",
-    "vite": "^8.0.0",
+    "vite": "8.0.8",
     "vite-tsconfig-paths": "^6.1.1",
     "vitest": "catalog:",
     "web-vitals": "^5.1.0"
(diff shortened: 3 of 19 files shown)


gronxb:
@codex review

gronxb:
Custom E2E for missing native launch wiring completed.

Scenario:
- iOS: temporarily changed AppDelegate release bundle URL to built-in `main.jsbundle` instead of `HotUpdater.bundleURL()`.
- Android: temporarily changed `MainApplication.getJSBundleFile()` to return `null` instead of `HotUpdater.getJSBundleFile(...)`.
- Used release binaries only.

Evidence:
- iOS runtime snapshot stayed on built-in/min bundle `019d96f0-3978-7000-8000-000000000000`, manifest bundle id stayed the same, runtime base URL was `null`.
- iOS bundle-store metadata had pending OTA staging bundle `019d9a6c-e72b-749a-be21-a9aa769987a2` with `verification_pending: true`.
- Android runtime snapshot stayed on min bundle `019d9a9a-f8fa-7000-8000-000000000000`, manifest bundle id stayed the same, and action result showed OTA install of `019d9a9f-2cc3-7320-b33d-3ad38de7e0fd`.
- Android bundle-store metadata had pending OTA staging bundle `019d9a9f-2cc3-7320-b33d-3ad38de7e0fd` with `verificationPending: true`.

Also verified PR checks are green after the ktlint fix.

PR #950, opened 2026-04-22
feat: add internal files directory retrieval in FileManagerService

HtSpChakradharCholleti:
This pull request updates the file storage location for bundles in the Android implementation of the HotUpdater module. The main change is switching from using the external files directory to the internal files directory for storing bundles, which improves data privacy and security. Additionally, the file system service interface and implementation are updated to support this change.

**File storage location update:**

* Changed the bundle storage location in `BundleFileStorageService` from the external files directory to the internal files directory by replacing usages of `getExternalFilesDir()` with `filesDir()`. [[1]](diffhunk://#diff-f1f20ea5ba3b8a5329f9c71c05c03a8ff7b3c4a38d93ec68c0e39cd91aa257f4L153-R153) [[2]](diffhunk://#diff-f1f20ea5ba3b8a5329f9c71c05c03a8ff7b3c4a38d93ec68c0e39cd91aa257f4L691-R691)

**File system service enhancements:**

* Added a new `filesDir()` method to the `FileSystemService` interface to provide access to the internal files directory.
* Implemented the new `filesDir()` method in `FileManagerService` to return the application's internal files directory.

PR #954, opened 2026-04-23
fix(android): align internal bundle store paths

gronxb:
## Summary
- Move the Android recovery bundle-store lookup from external app storage to internal `filesDir`.
- Update the Android Maestro store path from `/sdcard/Android/data/<appId>/files/bundle-store` to `/data/data/<appId>/files/bundle-store`.
- Read Android internal app files in Maestro through `run-as`, with the old direct shell read as a fallback.
- Build the Android E2E release APK as debuggable only when `HOT_UPDATER_E2E_DEBUGGABLE=true`, so `run-as` can inspect internal files during `pnpm -w e2e:maestro:android`.
- Remove the now-unused Android `FileSystemService.getExternalFilesDir()` API and update tests to use an internal files dir fake.

This is the follow-up PR for #950

### DIFF
--- a/e2e/maestro/flows/subflows/navigate-action-results.yaml
+++ b/e2e/maestro/flows/subflows/navigate-action-results.yaml
@@ -10,6 +10,19 @@ appId: ${APP_ID}
     waitToSettleTimeoutMs: 2000
 - waitForAnimationToEnd:
     timeout: 2000
+- scrollUntilVisible:
+    direction: DOWN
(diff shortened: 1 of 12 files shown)


gronxb:
@codex review

gronxb:
@codex Please review the code considering that this PR is a follow-up to PR #950.

gronxb:
@hot-updater-e2e android

PR #971, opened 2026-05-05
feat(cli): `hot-updater rollback <channel>`

dusty-cai:
Implements `rollback` command for #973 

---
"hot-updater": patch
---

feat(cli): add `rollback <channel>` command

Disables the most recent enabled bundle on a channel for each requested platform.

```
hot-updater rollback <channel> [-p ios|android] [-y] [--confirm-revert-to-binary] [--target <bundle-id>]
```

Behavior:

- **Read phase** loads up to two most-recent enabled bundles per (channel, platform) so the operator can see what would become active after rollback.
- **Validate phase** refuses with non-zero exit if any (channel, platform) would have **no** enabled bundles after the rollback unless `--confirm-revert-to-binary` is passed. The error message names both safe escape hatches in priority order: `-p <unaffected>` first, then `--confirm-revert-to-binary`.
- **Mutate phase** queues `updateBundle({ enabled: false })` for each target and commits once. Note: `DatabasePlugin.commitBundle` runs ops sequentially in the underlying provider, so atomicity across platforms is **not** guaranteed. The mutate is wrapped in a try/catch so a mid-commit throw still falls through to the verify phase.
- **Verify phase** re-reads each target. Distinguishes three states — disabled (success), still-enabled (failure), and gone (success: a deleted bundle satisfies the rollback intent). Surfaces partial-failure state explicitly with non-zero exit and per-platform `FAILED` lines naming the exact retry command, including a `--target <bundle-id>` flag for scoped retry.

Refuses to mutate without `-y` in a non-TTY shell. `onUnmount` is wrapped in its own try/catch so cleanup errors never mask the originating mutation error. Help text documents the four-phase contract and exit codes (0 = success, 1 = error, 2 = user-aborted).

### DIFF
--- a/.changeset/cli-rollback-command.md
+++ b/.changeset/cli-rollback-command.md
@@ -4,17 +4,17 @@
 
 feat(cli): add `rollback <channel>` command
 
-Disables the most recent enabled bundle on a channel for each requested platform. Operationally equivalent to revealing the second-most-recent bundle as the channel's active update.
+Disables the most recent enabled bundle on a channel for each requested platform.
 
 ```
-hot-updater rollback <channel> [-p ios|android] [-y] [--confirm-revert-to-binary]
+hot-updater rollback <channel> [-p ios|android] [-y] [--confirm-revert-to-binary] [--target <bundle-id>]
 ```
 
 Behavior:
 
 - **Read phase** loads up to two most-recent enabled bundles per (channel, platform) so the operator can see what would become active after rollback.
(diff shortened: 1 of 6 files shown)


gronxb (approved):

gronxb (changes requested):
I don’t think `--confirm-revert-to-binary` is necessary. Reverting to the embedded bundle is a natural case. So when the previous bundle cannot be found, there is no bundleId to print, so we should just inform the user that it will revert to the embedded bundle. Even if removing that flag results in zero items, it should still be allowed.

PR #972, opened 2026-05-05
Feature - CLI Bundle Commands

dusty-cai:
Implements `bundle` commands for #973 

---
"hot-updater": patch
---

feat(cli): add `bundle list/disable/enable` commands

Adds three subcommands under a new top-level `bundle` namespace:

- `hot-updater bundle list [-c <channel>] [-p <ios|android>] [--limit <n>]` — tabulated listing of bundles, most recent first. `--limit` validation uses commander's idiomatic `InvalidArgumentError` shape.
- `hot-updater bundle disable <bundle-id> [-y]` — disable a single bundle. Refuses to mutate without `-y` in a non-TTY shell. Re-reads the bundle after `commitBundle` and exits non-zero if the change did not take effect; treats a mid-flight deletion as success.
- `hot-updater bundle enable <bundle-id> [-y]` — re-enable a previously disabled bundle.

All three commands load config via `loadConfig(null)` (matching the `console` command's idiom) since they are not platform-scoped operations. They use the existing `DatabasePlugin` interface (`getBundles`, `getBundleById`, `updateBundle`, `commitBundle`), so they work against every supported provider with no plugin-side changes. The `--platform` option is the shared `platformCommandOption` already used by `deploy`. `onUnmount` is wrapped in its own try/catch so cleanup errors never mask the originating mutation error. Help text documents the read-mutate-verify contract and exit codes (0 = success, 1 = error, 2 = user-aborted).

### DIFF
--- a/packages/hot-updater/src/index.ts
+++ b/packages/hot-updater/src/index.ts
@@ -1,5 +1,9 @@
 #!/usr/bin/env node
-import { Command, Option } from "@commander-js/extra-typings";
+import {
+  Command,
+  InvalidArgumentError,
+  Option,
+} from "@commander-js/extra-typings";
 import type { AndroidNativeRunOptions } from "@hot-updater/android-helper";
 import type { IosNativeRunOptions } from "@hot-updater/apple-helper";
 import { banner, log, p } from "@hot-updater/cli-tools";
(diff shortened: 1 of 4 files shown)


packages/hot-updater/src/index.ts
```diff
@@ -82,6 +87,57 @@ channelCommand
...
   .action(handleSetChannel);
 
+const bundleCommand = program.command("bundle").description("Manage bundles");
+
+bundleCommand
+  .command("list")
+  .description("List bundles, most recent first")
+  .option("-c, --channel <channel>", "filter by channel")
+  .addOption(platformCommandOption)
+  .option(
```
gronxb, line 97: Could you add a `--json` option?

I think printing raw JSON would be a useful feature.

Issue #973, opened 2026-05-06
Feature - More CLI Commands


dusty-cai:
It's often easier to just type the command rather than run a command, open a browser, figure out what exists, then click a few more buttons. 

I propose to add:

```
hot-updater rollback <channel> [-p ios|android] [-y] [--confirm-revert-to-binary] [--target <bundle-id>]
hot-updater bundle list [-c <channel>] [-p <ios|android>] [--limit <n>]
hot-updater bundle disable <bundle-id> [-y]
hot-updater bundle enable <bundle-id> [-y]
hot-updater promote --source <channel> --target <channel> [--action move|copy] [--bundle-id <id>] [-y]
hot-updater deploy [-p ios|android|all] [-c <channel>] [-t <targetAppVersion>] ...     # -p all or omitted = both platforms, sequential, abort-on-first-failure
```

gronxb:
Sounds good ! 


I think it would be smoother to change `hot-updater promote` to `hot-updater channel promote`.

gronxb:
What you’re working on is valuable. I’ll merge this first, then refine the flags and prepare the documentation.

dusty-cai:
@gronxb: I've got changes prepped for the additional commands, I'll hold until you refine the promote command, much easier to change now than later

gronxb:
Since there is already a command called `hot-updater channel`, I meant that channel promotion was simply changed to use channel in front.

It looks like the functionality itself has not changed — only hot-updater promote was renamed to `hot-updater channel promote`. 

Keep going man

gronxb:
Thanks ! Released in v0.30.10

PR #980, opened 2026-05-06
feat(cli): add `hot-updater bundle promote` command

dusty-cai:
Move or copy a bundle to a different channel from the CLI, mirroring the console's Promote-to-Channel UI.

  hot-updater bundle promote <bundle-id> -t <target> [-a copy|move] [-y]

Lives under the `bundle` namespace alongside list/disable/enable because the noun being mutated is the bundle (its channel attribute, or a copy of it). Bundle id is positional; the bundle carries its own source channel, so no --source flag is needed.

--action copy (default) creates a new bundle id on the target channel and leaves the original in place -- CodePush-promote semantics. 
--action move updates the bundle's channel column without creating a new bundle (D1-only mutation; no R2 work).

Wraps the promoteBundle function exported from @hot-updater/cli-tools so the CLI and console use one implementation. Surfaces the underlying LEGACY_BUNDLE_ERROR and signing/storage configuration errors directly.

Pre-flight: rejects bundle-already-on-target, missing bundle id, empty target. Refuses to mutate without -y in a non-TTY shell.

Part of #973

### DIFF
--- a/.changeset/cli-promote-command.md
+++ b/.changeset/cli-promote-command.md
@@ -1,0 +1,18 @@
+---
+"hot-updater": patch
+---
+
+feat(cli): add `bundle promote` command
+
+Move or copy a bundle to a different channel from the CLI, mirroring the console's Promote-to-Channel UI.
+
+```
+hot-updater bundle promote <bundle-id> -t <target-channel> [-a copy|move] [-y]
+```
+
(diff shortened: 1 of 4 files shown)


dusty-cai:
@gronxb: I know you suggested `channel`, but after look at what the channel command does (just make local changes), it didn't feel like promote was the right place for it.

PR #985, opened 2026-05-07
Add R2 promote copy regression coverage

dusty-cai:
Revisited after #911 was merged.

Main now owns the storage profile split and R2 node download implementation through `NodeStorageProfile.downloadFile`. This PR now keeps that design and adds regression coverage for `bundle promote --action copy`: non-http storage URIs are downloaded through the node storage profile instead of `fetch`.

It also aligns the v0.85 example smoke config with the Cloudflare R2/D1 path used to verify the promote-copy scenario.

Validation:
- `pnpm lint`
- `pnpm exec vitest run --project=unit:default packages/cli-tools/src/promoteBundle.spec.ts plugins/cloudflare/src/r2Storage.spec.ts plugins/plugin-core/src/createStoragePlugin.spec.ts`
- `pnpm --filter @hot-updater/cli-tools test:type`
- GitHub Actions integration passed on head `b41d05c0`
- Vercel preview passed on head `b41d05c0`


plugins/cloudflare/src/r2Storage.ts
dusty-cai: This is what the new `download()` is bypassing... leaving it here, because it really doesn't support `getDownloadUrl()`, but it shouldn't be hit anymore.

gronxb (commented):
I’ll put this PR on hold. I understand why this issue is happening, and although the PR has grown a bit larger, I’m currently addressing it in this PR:

https://github.com/gronxb/hot-updater/pull/911

Once that PR is merged, a new `download` spec will be added. For now, I think it would be better to treat the Cloudflare promote issue as a known issue and simply make sure it throws errors properly.

gronxb:
This PR issue was also resolved in #911. Thanks!

PR #986, opened 2026-05-08
fix(cli): `deploy` defaults targetAppVersion from native files in non-interactive mode

dusty-cai:
## What

Wires `getDefaultTargetAppVersion`'s output into `deploy`'s non-interactive mode, so a CI invocation like `hot-updater deploy -c dev` no longer errors with "Target app version not found" when both `-t` and `-i` are omitted.

## Why

The default is already computed and used as the interactive prompt's placeholder/initialValue. Not threading it through to non-interactive mode meant every CI deploy had to either pass `-t` explicitly or scrape the version out of `package.json` — duplicating logic that `getDefaultTargetAppVersion` already centralizes.

## Resolution order

1. Explicit `-t <targetAppVersion>`
2. Interactive prompt (`-i`), with the auto-detected value as placeholder
3. Auto-detected default from native files (new)
4. Clear error if the native config is unreadable

The error path is now reachable only when the native files can't be parsed — pointing the operator at the actual problem (broken Info.plist / build.gradle) rather than asking them to retype the version they already shipped.

## Behavior change scope

- `hot-updater deploy -p ios -c dev -t 1.2.0` — unchanged (explicit `-t` still wins).
- `hot-updater deploy -p ios -c dev -i` — unchanged (interactive prompt still fires; placeholder still shows the auto-detected value).
- `hot-updater deploy -p ios -c dev` — was: error. Now: deploys at the binary's version.

## Tests

Added four scenarios to `packages/hot-updater/src/commands/deploy.spec.ts` covering the resolution order: non-interactive auto-fallback, non-interactive error-when-unreadable, explicit `-t` overriding the default, and interactive mode using the prompt with the auto-detected placeholder.

## Risk

Low. The change is an additional fallback rung; no existing successful invocation changes behavior. Operators who relied on the "no `-t`, no `-i`" path failing as a sanity check will instead get a successful deploy at the binary's version — which is what they would have typed anyway.

### DIFF
--- a/.changeset/cli-deploy-default-target-version.md
+++ b/.changeset/cli-deploy-default-target-version.md
@@ -1,0 +1,15 @@
+---
+"hot-updater": patch
+---
+
+fix(cli): `deploy` falls back to the auto-detected target app version in non-interactive mode
+
+Previously, running `hot-updater deploy` without `-t` and without `-i` errored with
+"Target app version not found", even though `getDefaultTargetAppVersion` had already
+extracted the version from the binary's native files (Info.plist for iOS, build.gradle
+for Android) for use as the interactive prompt's placeholder. CI deploys had to
+either pass `-t` explicitly or scrape the version out of package.json.
+
+Now the resolution order is: explicit `-t` → interactive prompt (with the auto-detected
+value as placeholder) → auto-detected default → clear error if the native config is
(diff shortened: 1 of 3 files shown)


gronxb (approved):
Cool ! Thanks !!

PR #999, opened 2026-05-15
chore(react-native): deprecate `updateMode` in `HotUpdater.wrap` 

gronxb:
`HotUpdater.init` instead of `updateMode: "manual"`


gronxb:
@codex review

PR #1003, opened 2026-05-16
chore: harden OTA management and bundle path handling

gronxb:
## Summary

- Disable self-hosted bundle management routes by default and require explicit opt-in when they are enabled.
- Keep `/version` always mounted for diagnostics; `routes` now only configures `updateCheck` and `bundles`.
- Document `routes.bundles: true` as the explicit opt-in that mounts bundle-management APIs so framework/reverse-proxy auth can protect `/hot-updater/api/*`.
- Harden Android and iOS OTA archive extraction and manifest-driven asset reuse against path traversal and sibling-prefix escapes.


gronxb:
@codex review

PR #1008, opened 2026-05-18
fix(hot-updater): match patch bases by semver compatibility

gronxb:
## Summary

- Match appVersion patch base discovery with update-check semver compatibility.
- Keep fingerprint patch base lookup on exact fingerprint matches.
- Add deploy coverage for non-exact compatible bases and overscan pagination, plus setup-based bsdiff update-check coverage that verifies patch metadata is returned instead of archive-only fallback.

## Root Cause

Automatic patch generation queried base bundles by exact `targetAppVersion`. A deploy targeting `1.1` skipped an existing `1.1.0` base, so no patch artifact was created. Later, an app reporting `1.1.0` could still receive the archive update, but there was no matching patch metadata to return.

## Validation

- `pnpm exec vitest run --project='unit:default' packages/hot-updater/src/commands/deploy.spec.ts`
- `pnpm exec vitest run --project='integration:cloudflare' plugins/cloudflare/worker/src/index.integration.spec.ts`
- `pnpm exec vitest run --project='integration:default' plugins/aws/lambda/runtime.container.integration.spec.ts`
- `pnpm -w test:integration`
- `pnpm -w lint`

### DIFF
--- a/.changeset/hot-birds-compare.md
+++ b/.changeset/hot-birds-compare.md
@@ -1,0 +1,5 @@
+---
+"hot-updater": patch
+---
+
+fix(hot-updater): match patch bases by semver compatibility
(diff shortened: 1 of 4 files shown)


gronxb:
@codex review

PR #1013, opened 2026-05-19
[codex] Limit manifest asset upload concurrency

gronxb:
## Summary

- Limit deploy manifest asset uploads to 8 concurrent storage requests.
- Add a regression test that verifies concurrent asset uploads stay bounded.

## Root Cause

File-level manifests can produce hundreds of asset uploads during deploy. The deploy path previously sent all asset uploads through an unbounded `Promise.all`, which could fan out hundreds of Supabase Storage requests at once. Supabase Storage stores object metadata through its database layer, so that burst can exhaust the project connection limit and fail with `The database has reached its maximum number of connections`.

## Impact

Deploys with many manifest assets now keep storage upload concurrency bounded while preserving manifest-driven diff update behavior.

## Validation

- `pnpm -w exec vitest run packages/hot-updater/src/commands/deploy.spec.ts`
- `pnpm --filter hot-updater test:type`
- `pnpm lint`

## Not Tested

- Live Supabase deploy against a low-connection project.

### DIFF
--- a/packages/hot-updater/src/commands/deploy.spec.ts
+++ b/packages/hot-updater/src/commands/deploy.spec.ts
@@ -200,6 +200,7 @@ vi.mock("./console", () => ({
 }));
 
 import fs from "fs";
+import path from "path";
 
 import type { Bundle, DatabasePlugin } from "@hot-updater/plugin-core";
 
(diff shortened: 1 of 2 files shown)


gronxb:
Replaced by #1014 with the repository branch/title convention and changeset included.

Issue #1010, opened 2026-05-18
Android `hot_updater_channel` string should be generated with `translatable="false"`


MidnightCoke:
### Precheck

- [x] Infrastructure may have changed. Did you try `hot-updater doctor`? (https://hot-updater.dev/docs/guides/doctor)
- [x] Is your Node version v20 or higher?
- [x] Are you aware that the update will only work properly if you `hot-updater deploy` after building the native app?

### Name, Version of hot-updater

0.31.2

### Version of react-native

0.83.6

### Platforms

Android

### Architecture

New Architecture(Fabric)

### What provider are you seeing the problem on? (multiple)

S3 & Lambda

### Expo App

- [x] I am using Expo
- [ ] Project is managed with Expo [CNG](https://docs.expo.dev/workflow/continuous-native-generation/)(android, ios directories are included in `.gitignore`)

### Bundler

metro(hermes)

## Summary

On Android release builds, the Hot Updater Expo config plugin generates the update
channel as a normal Android string resource:

```xml
<string name="hot_updater_channel" moduleConfig="true">production</string>
```

Because the string is not marked as non-translatable, Android/Google Play resource
localization can translate the channel value. On a Turkish device, our runtime
diagnostics showed the configured channel `production` becoming `üretim`.

This makes the app check the wrong Hot Updater channel and return `UP_TO_DATE`,
even when an update exists on the intended `production` channel.

## Environment

- `hot-updater`: `0.31.2`
- `@hot-updater/react-native`: `0.31.2`
- `@hot-updater/aws`: `0.31.2`
- `@hot-updater/expo`: `0.31.2`
- `@hot-updater/sentry-plugin`: `0.31.2`
- React Native / Expo Android release build
- Update strategy: `appVersion`
- Configured channel: `production`
- Device locale: Turkish

## Relevant Config

`app.json`:

```json
[
  "@hot-updater/react-native",
  {
    "channel": "production"
  }
]
```

`hot-updater.config.ts`:

```ts
export default defineConfig({
  updateStrategy: 'appVersion',
  storage: s3Storage(commonOptions),
  database: s3Database(commonOptions),
});
```

Generated Android resource:

```xml
<string name="hot_updater_channel" moduleConfig="true">production</string>
```

## Actual Runtime Behavior

The release build logged:

```text
[HotUpdater] app ready
{
  "result": { "status": "STABLE" },
  "diagnostics": {
    "appVersion": "1.2.0",
[...]

PR #1017, opened 2026-05-19
fix(android): avoid localized native config values

gronxb:
## Summary

- Move Android Hot Updater native config writes from `strings.xml` to `AndroidManifest.xml` `<meta-data>`.
- Read manifest metadata first and keep legacy `strings.xml` values as fallback.
- Update Expo plugin, CLI XML mutation commands, doctor/signing validation, and Android native runtime reads.

Reference: https://github.com/gronxb/hot-updater/issues/1010#issuecomment-4486898180


gronxb:
@codex review

PR #1035, opened 2026-05-28
fix(cloudflare): inline IN-list values to avoid D1 100-bind cap

### LINKED ISSUE #1034: d1WorkerDatabase: IN(...) lists overrun D1's 100-bind cap → SQLITE_ERROR
# `d1WorkerDatabase`: `IN(...)` lists overrun D1's 100-bind cap → SQLITE_ERROR

## Summary

`plugins/cloudflare/src/cloudflareWorkerDatabase.ts` builds three `IN (...)`
clauses by expanding the array into `?, ?, ?, ...` and pushing each value
into `.bind(...)`. D1 caps prepared statements at **100 bound parameters**,
so any of these queries fail with `D1_ERROR: too many SQL variables ...
SQLITE_ERROR` once the array crosses ~95 entries.

Every `/api/check-update` request 5xx's once it triggers. Logs:

```
Hot Updater handler error:
  Error: D1_ERROR: too many SQL variables at offset 386: SQLITE_ERROR
```

## Affected call sites

- `buildWhereClause` — `id.in` (line ~114)
- `buildWhereClause` — `targetAppVersionIn` (line ~161)
- `getPatchMap` — `bundle_id IN (?, ?, ...)` (line ~338)

The `targetAppVersionIn` path is the easiest to trip in production: the
update-check flow first calls `getTargetAppVersionsForUpdateInfo` to fetch
every distinct `target_app_version`, filters in JS via semver, then passes
the whole filtered list as `targetAppVersionIn`. As soon as a project
accumulates ~100 distinct compatible range strings, every check-update
errors.

## Reproduction

Insert 200 bundles with distinct `target_app_version` ranges all
compatible with `1.0.0`, then call `getUpdateInfo` for an iOS client at
`1.0.0`:

```ts
for (let i = 0; i < 200; i++) {
  await db.commitBundle({
    changedSets: [{
      operation: "insert",
      data: { /* ... target_app_version: `>=0.${i}.0`, ... */ }
    }]
  });
}
// → throws D1_ERROR: too many SQL variables
```

## Proposed fix

The values in these IN-lists are written by our own backend
(`bundle.id` is a uuid; `target_app_version` is the semver range the
uploader supplied), so inline them as escaped SQL string literals
instead of binds. The number of binds in the prepared statement stays
constant regardless of list size.

A helper:

```ts
function toSqlLiteral(value: unknown): string | null {
  if (typeof value !== "string") return null;
  if (/[\n\r\0\\]/.test(value)) return null;
  return `'${value.replace(/'/g, "''")}'`;
}
```

is used at the three affected sites. Empty lists short-circuit to
`1 = 0` (since `IN ()` is a SQLite syntax error).

Happy to send a PR with the change + a vitest case that asserts 200
distinct target_app_versions can be queried without error. Want me to
open it?

## Environment

- `@hot-updater/cloudflare` version: (whichever the user is on, e.g. 0.x.x)
- Runtime: Cloudflare Workers + D1
- Affected since: as long as the IN-expansion has been in the file
  (predates `targetAppVersionIn` becoming the hot path)

KeyEu:
# fix(cloudflare): inline IN-list values to avoid D1 100-bind cap

## Summary

`plugins/cloudflare/src/cloudflareWorkerDatabase.ts` expanded three
`IN (...)` clauses as `?, ?, ?, ...` and pushed each value into
`.bind(...)`:

- `buildWhereClause` — `id.in`
- `buildWhereClause` — `targetAppVersionIn`
- `getPatchMap` — `bundle_id IN (...)`

D1 caps prepared statements at **100 bound parameters**, so any of
these queries fail with `D1_ERROR: too many SQL variables ...
SQLITE_ERROR` once the array crosses ~95 entries.

The hot path is `targetAppVersionIn`: the update-check flow first calls
`getTargetAppVersionsForUpdateInfo` to fetch every distinct
`target_app_version`, semver-filters in JS, then hands the filtered
list to `getBundlesByTargetAppVersions` → `queryBundlesForUpdateInfo`
→ `buildWhereClause`. Once a project accumulates ~100 distinct
compatible range strings, every `/api/check-update` request 5xx's.

Repro logs:

```
Hot Updater handler error:
  Error: D1_ERROR: too many SQL variables at offset 386: SQLITE_ERROR
```

## Approach

The values in these IN-lists are produced by our own backend (bundle
ids are uuids; target_app_version strings come from the uploader), so
they're safe to inline as escaped SQL string literals via a small
`toSqlLiteral` helper:

```ts
function toSqlLiteral(value: unknown): string | null {
  if (typeof value !== "string") return null;
  if (/[\n\r\0\\]/.test(value)) return null;
  return `'${value.replace(/'/g, "''")}'`;
}
```

- Single quotes are doubled per the SQL standard.
- Newline / NUL / backslash are rejected defensively so a future
  malformed row can't break out of the literal.
- Empty lists short-circuit to `1 = 0` (`IN ()` is a SQLite syntax
  error).
- The 5 binds in `getUpdateInfo`'s outer CTE are unchanged — only the
  IN-list values move from bind to inline.

Also: `getTargetAppVersionsForUpdateInfo` gains `ORDER BY MAX(id) DESC`
so the returned list is deterministic newest-first. This has no
functional impact today (the downstream filter is order-insensitive)
but removes reliance on SQLite's physical row order.

## Test coverage

`d1Database.spec.ts` was relying on `?` count to parse IN-list values
out of mock SQL. Updated the spec mock to handle both forms (bound and
inlined). Added two regression tests:

1. `getUpdateInfo` with 200 distinct `target_app_version` values
   (exceeds the old cap by 2×).
2. `getBundles` returning 200 rows (drives `getPatchMap` past the cap).

Both pass; full suite (114 tests) green.

## Files changed

- `plugins/cloudflare/src/cloudflareWorkerDatabase.ts` — `toSqlLiteral`
  helper + three IN-clause sites + ORDER BY tweak.
- `plugins/cloudflare/src/d1Database.spec.ts` — mock parses both forms;
  two new regression tests.

## Notes

I considered a schema-level fix (store resolved `min_app_version` /
`max_app_version` per bundle so the read path can use a single
`BETWEEN` instead of fetch-all + semver-in-JS + IN-list). That's the
architecturally cleaner approach but it's a bigger lift (migration +
backfill + uploader change + a new semver-range resolver becomes
load-bearing) and out of scope for fixing a live incident. Happy to
follow up with that in a separate PR if it's interesting.

Closes #1034.

### DIFF
--- a/plugins/cloudflare/src/cloudflareWorkerDatabase.ts
+++ b/plugins/cloudflare/src/cloudflareWorkerDatabase.ts
@@ -1,3 +1,25 @@
+// plugins/cloudflare/src/cloudflareWorkerDatabase.ts
+//
+// Drop-in replacement for the official Hot Updater Cloudflare worker
+// database plugin (https://github.com/gronxb/hot-updater).
+//
+// Fix: D1_ERROR "too many SQL variables ... SQLITE_ERROR"
+//
+// The upstream `buildWhereClause` and `getPatchMap` expand IN-lists as
+// `?, ?, ?, ...`. D1 caps prepared statements at 100 bound parameters,
+// so as soon as `targetAppVersionIn` / `id.in` / `bundleIds` cross ~95,
+// every request 5xxs. Values for these IN-lists come from our own DB
+// (bundle ids, target_app_version strings), so we inline them as
+// escaped SQL literals instead of binds. Empty lists short-circuit
+// (`IN ()` is a SQLite syntax error).
+//
+// Also: `getTargetAppVersionsForUpdateInfo` now sorts by `MAX(id) DESC`
+// so the returned target_app_version list is deterministic — newest
+// activity first — without needing an extra JS sort.
+//
+// No semantics change beyond those three points. Everything else is
+// byte-identical to upstream.
+
 import {
   DEFAULT_ROLLOUT_COHORT_COUNT,
   getAssetBaseStorageUri,
@@ -82,6 +104,16 @@ interface D1WorkerBundlePatchRow {
   order_index: number | null;
 }
 
+// Escape `value` as a SQL string literal, or return null if it contains
+// a character we refuse to inline (newline / NUL / backslash). Doubled
+// single-quote per the SQL standard. Only used for IN-list inlining;
+// every other parameter still goes through .bind().
+function toSqlLiteral(value: unknown): string | null {
+  if (typeof value !== "string") return null;
+  if (/[\n\r\0\\]/.test(value)) return null;
+  return `'${value.replace(/'/g, "''")}'`;
+}
+
 function buildWhereClause(
   conditions: QueryConditions | undefined,
 ): BuildQueryResult {
@@ -111,8 +143,14 @@ function buildWhereClause(
     if (conditions.id.in.length === 0) {
       clauses.push("1 = 0");
     } else {
-      clauses.push(`id IN (${conditions.id.in.map(() => "?").join(", ")})`);
-      params.push(...conditions.id.in);
+      const literals = conditions.id.in
+        .map(toSqlLiteral)
+        .filter((v): v is string => v !== null);
+      if (literals.length === 0) {
+        clauses.push("1 = 0");
+      } else {
+        clauses.push(`id IN (${literals.join(", ")})`);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


gronxb:
Thanks again for the clear report and initial fix. I opened #1036 with the same failure covered, but using D1's `json_each(?)` pattern so the values stay bound instead of being inlined into SQL. I also included you as co-author on the commit.

Closing this in favor of #1036 so we can merge the maintained branch.

PR #1038, opened 2026-05-29
fix(supabase): avoid PostgREST overflow during update checks

gronxb:
## Summary

- route Supabase `getUpdateInfo` through the existing SQL/RPC lookup functions
- keep app-version compatibility filtering in TypeScript, then pass the filtered ranges into the RPC query
- keep the high-cardinality getUpdateInfo integration fixture, but give that deliberate large-input case a 120s timeout
- remove the stale Cloudflare worker update lookup helper that still used placeholder-expanded D1 `IN` queries
- add a patch changeset for `@hot-updater/supabase`

## Root cause

The Supabase update-check path used PostgREST REST filters to fetch matching bundles and then hydrate patch rows with an `IN` filter over bundle ids. With many distinct compatible `target_app_version` ranges, that REST query can become oversized and fail in Supabase Edge/PostgREST.

This keeps the update lookup database-side via the existing Supabase RPC functions instead of splitting one logical update check into client-managed REST chunks. Cloudflare's active worker database path already uses JSON array binding via `json_each(?)`; the old unused worker helper was removed so future fixes do not follow the stale placeholder-expanded query shape.

## Validation

  - `pnpm -w build`
  - `pnpm -w test:type`
  - `pnpm -w lint`
  - `pnpm -w test`
  - `pnpm -w test:integration`
- `pnpm --filter @hot-updater/cloudflare test:type`
- `pnpm --filter @hot-updater/cloudflare build`
- `pnpm exec vitest run plugins/cloudflare/src/cloudflareWorkerDatabase.spec.ts --project=unit:default`


gronxb:
@codex review

PR #1044, opened 2026-06-10
fix(cli-tools): upload brotli assets when copying bundles

debugtheworldbot:
## Summary

Fixes copied bundle promotion so Hermes bundle assets are uploaded with the Brotli artifact name expected by manifest-driven update checks.

When `promoteBundle` copies a bundle to another channel, it rewrites the manifest and re-uploads the archive, manifest, and manifest assets. The server resolves `index.<platform>.bundle` changed assets as `index.<platform>.bundle.br` and marks them with `compression: "br"`, matching deploy behavior. Copy promotion previously uploaded the extracted `index.<platform>.bundle` without the `.br` suffix, which left update-check responses pointing at a missing object.

This change Brotli-compresses copied manifest bundle assets before upload and stores them as `index.<platform>.bundle.br`, while leaving the archive and manifest paths unchanged.

## Validation

- `pnpm --filter @hot-updater/core --filter @hot-updater/plugin-core build`
- `pnpm --filter @hot-updater/js build`
- `pnpm exec vitest run packages/cli-tools/src/promoteBundle.spec.ts`
- `pnpm --filter @hot-updater/cli-tools test:type`
- `pnpm exec oxfmt --check packages/cli-tools/src/promoteBundle.ts packages/cli-tools/src/promoteBundle.spec.ts`

### DIFF
--- a/packages/cli-tools/src/promoteBundle.ts
+++ b/packages/cli-tools/src/promoteBundle.ts
@@ -1,8 +1,10 @@
 import crypto from "node:crypto";
+import { createReadStream, createWriteStream } from "node:fs";
 import fs from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
-import { brotliCompressSync, brotliDecompressSync } from "node:zlib";
+import { pipeline } from "node:stream/promises";
+import { createBrotliCompress, brotliDecompressSync } from "node:zlib";
 
 import {
   getManifestFileHash,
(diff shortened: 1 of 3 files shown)


gronxb:
@codex review

packages/cli-tools/src/promoteBundle.ts
```diff
@@ -78,6 +80,52 @@ const getRelativeStorageDir = (relativePath: string) => {
   return dirname === "." ? "" : dirname;
 };
 
+const isBrotliManifestBundleAsset = (relativePath: string) =>
```
gronxb: The bundle file is not guaranteed to be named index; it should match the pattern `*.(ios|android).bundle`.

gronxb: how about naming isBundleAsset

gronxb (changes requested):

gronxb (approved):

PR #1045, opened 2026-06-10
fix(console): make selected bundle deletion observable

gronxb:
## What changed
- Replaced the selected-bundle bulk delete alert with a shadcn-composed progress dialog.
- Shows each selected bundle in a table log with queued, deleting, deleted, and failed states.
- Keeps the modal locked during active deletion, then exposes Close or Retry failed actions after completion.
- Keeps failed bundle IDs selected in the table so users can act on the remaining failures.

## Why
Bulk deletion runs sequentially and can take a long time per bundle. The previous UI only showed a generic pending button, so users could not tell which bundle was being deleted or whether individual items succeeded.

## Validation
- pnpm --filter @hot-updater/console test -- SelectedBundlesDeleteDialog.spec.tsx
- pnpm --filter @hot-updater/console test:type
- pnpm lint
- pnpm --filter @hot-updater/console build
- Browser smoke test at http://localhost:3002/: selected two bundles, opened the delete dialog, completed deletion, and verified the modal showed 2 of 2 finished with both rows marked Deleted and Close available.

## Not tested
- Full workspace pnpm -w test was not run.

### DIFF
--- a/packages/console/src/components/features/bundles/BundlesTable.tsx
+++ b/packages/console/src/components/features/bundles/BundlesTable.tsx
@@ -14,7 +14,6 @@ import {
   Trash2,
 } from "lucide-react";
 import { Fragment, useState } from "react";
-import { toast } from "sonner";
 
 import { BundleIdDisplay } from "@/components/BundleIdDisplay";
 import { ChannelBadge } from "@/components/ChannelBadge";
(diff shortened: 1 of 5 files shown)


gronxb:
@codex review

PR #1047, opened 2026-06-11
fix(server): keep update bundle reuse in helpers

gronxb:
## Summary

- keep per-request update bundle reuse in the server/core + plugin-core helper layer
- add `resolveUpdateInfoFromBundles` so direct manifest update checks seed request-local bundle state without provider code handling identity-map internals
- keep `seedRequestUpdateBundles` internal to plugin-core modules while exposing only the server-readable seed accessor

## Context

#1046 added a per-request identity map, but the provider `getUpdateInfo` path still selected bundles from already-loaded manifest lists and returned only `UpdateInfo`. Server artifact resolution then had no object identity seed for those selected bundles, so it could still re-read target/current bundles by id.

This PR moves that missing bridge into plugin-core helpers instead of requiring provider authors to think about request-local cache seeding. Blob/direct lookup providers stay on direct manifest reads, and the server consumes request seeds consistently when resolving update artifacts.

## Verification

- `pnpm exec vitest run packages/server/src/db/pluginCore.spec.ts packages/server/src/db/requestBundleIdentityMap.spec.ts plugins/plugin-core/src/createBlobDatabasePlugin.spec.ts plugins/plugin-core/src/createDatabasePlugin.spec.ts plugins/plugin-core/src/createDatabasePluginGetUpdateInfo.spec.ts`
- `pnpm --filter @hot-updater/plugin-core test:type`
- `pnpm --filter @hot-updater/server test:type`
- `pnpm exec oxlint packages/server/src/db/pluginCore.ts packages/server/src/db/pluginCore.spec.ts packages/server/src/db/requestBundleIdentityMap.ts plugins/plugin-core/src/createBlobDatabasePlugin.ts plugins/plugin-core/src/createBlobDatabasePlugin.spec.ts plugins/plugin-core/src/createDatabasePluginGetUpdateInfo.ts plugins/plugin-core/src/index.ts plugins/plugin-core/src/requestUpdateBundleState.ts plugins/plugin-core/src/resolveUpdateInfoFromBundles.ts`
- `pnpm --filter @hot-updater/plugin-core build`
- `pnpm --filter @hot-updater/server build`
- `pnpm --filter @hot-updater/aws build`

### DIFF
--- a/packages/server/src/db/pluginCore.spec.ts
+++ b/packages/server/src/db/pluginCore.spec.ts
@@ -4,6 +4,7 @@ import type {
   DatabasePlugin,
   RequestEnvContext,
 } from "@hot-updater/plugin-core";
+import { createDatabasePluginGetUpdateInfo } from "@hot-updater/plugin-core";
 import { describe, expect, it, vi } from "vitest";
 
 import { createPluginDatabaseCore } from "./pluginCore";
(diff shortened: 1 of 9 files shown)


gronxb:
Verification update

- Confirmed #1028 no longer includes server/plugin-core files: `gh pr view 1028 --json files --jq ...` returned `0` matching files, and `origin/main...origin/analysis/issue-1026` has no `packages/server/src/db` or `plugins/plugin-core/src` diff.
- Confirmed this PR contains the migrated server/helper changes only: 9 files under `packages/server/src/db` and `plugins/plugin-core/src`.
- Confirmed the pushed #1028 head is `db327321`, which reverts the two server-fix commits from that PR scope without rewriting history.

AWS timeout-focused checks on `codex/request-update-identity-map`:

- `pnpm exec vitest run plugins/aws/lambda/updateCheck.memory.spec.ts --testTimeout=10000` passed in `real 1.17s`; Vitest test duration was `403ms`.
- `pnpm exec vitest bench --project unit:default plugins/aws/lambda/updateCheck.bench.ts --run --testTimeout=10000` passed in `real 2.18s`; app-version route max `1.1686ms`, fingerprint route max `0.3747ms`.
- `pnpm --filter @hot-updater/aws build` passed in about `2.2s`, including the packaged Lambda entrypoint build.

Live AWS Lambda@Edge was not invoked; this verifies the local packaged AWS update-check surface and direct blob/helper path against a 10s test timeout.

gronxb:
@codex review

PR #1051, opened 2026-06-11
feat(plugin-core): add request-scoped bundle identity map

gronxb:
## Summary

- Remove derived blob management index reads/writes from the AWS blob database path so stale `_index` snapshots cannot hide canonical bundles.
- Limit AWS/S3 bundle discovery to canonical `update.json` manifest paths instead of recursively listing uploaded assets.
- Keep same-process bundle deletes from forcing an immediate storage reload after the bundle was already deleted locally.
- Add a request-scoped bundle unit-of-work in plugin-core so provider implementations do not need to manage identity-map caching.
- Guarantee same-request bundle identity for `getBundleById -> updateBundle -> getBundleById`, make `getBundles` reflect pending update/delete state with query-aware overlays, and clear pending state after commit.
- Reuse the request identity map during server update-info artifact resolution so manifest artifact lookup does not add duplicate bundle DB reads.
- Clear provider-local pending blob state during canonical reloads so another plugin instance's committed S3 manifest update is visible through the canonical path.

## Context

The original AWS optimization still stands: canonical update manifests are the source of truth, and the derived blob management index is removed from deploy/reset flows because stale incremental state can hide real bundles.

This branch also replaces the PR-local read cache shape with a core unit-of-work / identity-map boundary. Request contexts own bundle identity, pending mutations are tracked centrally, and plugin authors continue to implement only provider reads/writes. No-context read calls no longer persist identity entries across logical requests; no-context mutation staging is kept only until `commitBundle()` for existing CLI-style flows.

AWS full E2E still does not match the 20-minute provider baseline because the dashboard serializes AWS shards to avoid historical CloudFront check-update failures under concurrent full-platform load. Within that constraint, the previous successful AWS full run completed in `1h35m0s`, down from the earlier AWS success baseline of `1h51m0s`.

## Validation

Current local validation after the latest commit:

- `pnpm exec vitest run --project=unit:default plugins/plugin-core/src/createBlobDatabasePlugin.spec.ts plugins/aws/src/s3Database.spec.ts`
- `pnpm exec vitest run plugins/plugin-core/src/bundleUnitOfWork.spec.ts plugins/plugin-core/src/createDatabasePlugin.spec.ts plugins/plugin-core/src/createBlobDatabasePlugin.spec.ts packages/server/src/db/pluginCore.spec.ts packages/server/src/db/createBundleDiff.spec.ts`
- `pnpm exec vitest run packages/server/src/db/pluginCore.spec.ts packages/server/src/db/createBundleDiff.spec.ts`
- `pnpm --filter @hot-updater/plugin-core test:type`
- `pnpm --filter @hot-updater/aws test:type`
- `pnpm --filter @hot-updater/server test:type`
- `pnpm --filter @hot-updater/plugin-core build`
- `pnpm --filter @hot-updater/aws build`
- `pnpm --filter @hot-updater/server build`
- `pnpm exec oxlint <changed files>`
- `pnpm exec oxfmt --check <changed files>`
- `git diff --check`

Previous E2E evidence retained from the AWS optimization pass:

- `hot-updater-agent verify -platform full -profile aws -env-target examples/v0.85.0/.env.hotupdater -tail 240` (`job-20260611204122-kcqhhj`, success, `1h35m0s`)

Not rerun after the latest identity-map commit:

- AWS full hot-updater-agent E2E.

## Notes

- This PR was reopened as #1051 after the accidental direct main merge was reverted.
- AWS full wall-clock parity with Cloudflare/Firebase/Supabase is not achieved by AWS provider changes alone while AWS full shards remain serialized.
- Reintroducing a persisted bundle-id index could reduce fresh `bundle delete <id>` lookups, but it brings back the stale/missing derived-index risk this PR removes.


gronxb:
@codex review

Issue #849, opened 2026-03-01
How do I update the server on Cloudflare?


bbhxwl:
I updated the SDK of the app today, but I don't know how to update it on https://hot-updater.dev/docs/managed/cloudflare/?

   "@hot-updater/cloudflare": "^0.25.13",
    "@hot-updater/expo": "^0.25.13",
    "@hot-updater/react-native": "^0.25.13",

gronxb:
you can try `(pnpm|npx|yarn) hot-updater init`

bbhxwl:
> you can try `(pnpm|npx|yarn) hot-updater init`

Isn't this a reinitialization? I have deployed it before, and after updating the latest dependencies, I want to update the worker

gronxb:
It was designed so that running `hot-updater init` again triggers a redeployment. Any required migrations are also performed only as much as needed, so you can simply run `hot-updater init` again.


Therefore, running `hot-updater init` updates the worker to the latest version again.

PR #1060, opened 2026-06-18
feat(hot-updater): add `--provider` and `--build` flags to `init`

### LINKED ISSUE #1059: Non-interactive `init` (`--provider`/`--build`) for scripted worker redeploys
### Summary

Add optional `--provider` and `--build` flags to `hot-updater init` so the worker redeploy can run non-interactively (CI, scripted SDK upgrades).

### Background

In #849 the answer to "how do I update the worker after bumping the SDK" is to re-run `hot-updater init` — it reuses the existing `.env.hotupdater` values and applies only the migrations that are needed. That's a clean design and works well interactively.

The problem is that `init` always prompts for the build plugin and the provider (the two `p.select` calls at the top of `init`). So the redeploy path can't run unattended — e.g. a CI job that bumps `@hot-updater/cloudflare` and redeploys, or any scripted upgrade.

Everything downstream of those two selects already short-circuits on existing `.env.hotupdater` values (account id, bucket, D1, worker name) and redeploys the worker unconditionally, and `ensureInstallPackages` is a no-op when the packages are already installed. So the only thing forcing interactivity on a re-run is the build/provider selection.

### Proposal

```
hot-updater init --provider cloudflare --build expo
```

- When a flag is passed, skip that select; validate the value against the known set and exit with a clear error otherwise.
- When omitted, prompt exactly as today — fully backward compatible.

Combined with an existing `.env.hotupdater`, this makes `init` a non-interactive "redeploy worker + apply pending migrations" command, which is the #849 workflow in a form you can put in CI.

I'll open a PR for this.

d4mr:
## What

Adds two optional flags to `hot-updater init`:

```
hot-updater init --provider cloudflare --build expo
```

`--provider` (cloudflare, aws, supabase, firebase) and `--build` (bare, rock, expo) pre-answer the two prompts that `init` always shows. When a flag is omitted the prompt is shown exactly as before, so this is fully backward compatible. Values are validated with commander's `.choices()`, matching the existing `bundle promote --action` option.

Closes #1059. Background in #849.

## Why

Per #849, re-running `hot-updater init` is the intended way to redeploy the worker after bumping the SDK. On Cloudflare that path is idempotent: it reuses the existing `.env.hotupdater` values and applies only the migrations that are needed. The only thing forcing interaction on a re-run is the build plugin and provider selects at the top of `init`. These flags remove them, so the redeploy can run unattended (for example a CI job that bumps `@hot-updater/cloudflare` and redeploys).

## Scope, and where I am unsure

I want to be upfront that this is a small, pragmatic change that helps my own workflow, and I do not think it is the complete answer to non-interactive `init`:

- It only removes the two top-level prompts. It does not make `init` guaranteed non-interactive. On Cloudflare, any value missing from `.env.hotupdater` still falls back to its own prompt (account, R2 bucket, D1, worker name), and worker-deploy auth still uses the wrangler OAuth session, so a fully headless run also needs a valid wrangler login.
- The reuse-existing-env behavior that turns a re-run into a redeploy looks, as far as I can tell, mostly like a Cloudflare property. AWS and Supabase `runInit` prompt on every run and do not read existing `.env.hotupdater`, so for those providers these flags only skip the two selects and the rest stays interactive. Firebase appears to read the env partially.

So this is best read as a convenience that makes the Cloudflare redeploy flow scriptable, not a full `--non-interactive` mode. A complete version (error instead of prompt when a required value is missing, consistent env reuse across providers) would be a larger change across each provider's iac. Happy to follow up on that if you think it is worth doing, or to adjust the framing/flag names to your preference.

## Testing

- `pnpm --filter hot-updater test:type` passes.
- `oxfmt` and `oxlint` clean on the changed files.
- Built and smoke-tested the CLI: both flags appear in `init --help`, and an invalid value is rejected by commander (`--provider bogus` reports the allowed choices).

A changeset is included (`hot-updater` minor).

---

This PR was assisted by Claude Code.

### DIFF
--- a/packages/hot-updater/src/index.ts
+++ b/packages/hot-updater/src/index.ts
@@ -79,7 +79,22 @@ program
   .description(banner(version))
   .version(version as string);
 
-program.command("init").description("Initialize Hot Updater").action(init);
+program
+  .command("init")
+  .description("Initialize Hot Updater")
+  .addOption(
+    new Option(
+      "--provider <provider>",
+      "provider to use; skips the prompt",
+    ).choices(["cloudflare", "aws", "supabase", "firebase"]),
+  )
+  .addOption(
+    new Option(
+      "--build <plugin>",
+      "build plugin to use; skips the prompt",
+    ).choices(["bare", "rock", "expo"]),
+  )
+  .action((options) => init(options));
 
 program
   .command("doctor")
--- a/.changeset/init-noninteractive-flags.md
+++ b/.changeset/init-noninteractive-flags.md
@@ -1,0 +1,15 @@
+---
+"hot-updater": minor
+---
+
+feat(hot-updater): add `--provider` and `--build` flags to `init`
+
+`hot-updater init` always prompts for the build plugin and the provider. These optional flags pre-answer those two prompts so `init` can run without interaction:
+
+```
+hot-updater init --provider cloudflare --build expo
+```
+
+When a flag is omitted, the prompt is shown as before. Values are validated against the known choices.
+
+This is aimed at the Cloudflare redeploy flow described in #849: with a populated `.env.hotupdater`, re-running `init` redeploys the worker and applies pending migrations, and these flags remove the two prompts that otherwise block it from running unattended. Providers still prompt for any value that is not already present in `.env.hotupdater`.
--- a/packages/hot-updater/src/commands/init.ts
+++ b/packages/hot-updater/src/commands/init.ts
@@ -11,6 +11,44 @@ const REQUIRED_PACKAGES = {
   devDependencies: ["dotenv"],
 };
 
+interface BuildPluginChoice {
+  name: string;
+  label: string;
+  hint?: string;
+  dependencies: string[];
+  devDependencies: string[];
+}
+
+const BUILD_PLUGINS: Record<"bare" | "rock" | "expo", BuildPluginChoice> = {
+  bare: {
+    name: "bare",
+    label: "Bare",
+    hint: "React Native CLI",
+    dependencies: [],
+    devDependencies: ["@hot-updater/bare"],
+  },
+  rock: {
+    name: "rock",
+    label: "Rock",
+    hint: "React Native Enterprise Framework by Callstack",
+    dependencies: [],
+    devDependencies: ["@hot-updater/rock"],
+  },
+  expo: {
+    name: "expo",
+    label: "Expo",
+    dependencies: [],
+    devDependencies: ["@hot-updater/expo"],
+  },
+};
+
+const PROVIDER_LABELS = {
+  cloudflare: "Cloudflare D1 + R2 + Worker",
+  aws: "AWS S3 + Lambda@Edge",
+  supabase: "Supabase",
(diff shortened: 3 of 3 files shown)


packages/hot-updater/src/index.ts
```diff
@@ -79,7 +79,22 @@ program
...
+    new Option(
+      "--provider <provider>",
+      "provider to use; skips the prompt",
+    ).choices(["cloudflare", "aws", "supabase", "firebase"]),
+  )
+  .addOption(
+    new Option(
+      "--build <plugin>",
+      "build plugin to use; skips the prompt",
+    ).choices(["bare", "rock", "expo"]),
```
gronxb, line 95:
```suggestion
    ).choices(Object.keys(BUILD_PLUGINS)),
```

packages/hot-updater/src/index.ts
```diff
@@ -79,7 +79,22 @@ program
...
 
-program.command("init").description("Initialize Hot Updater").action(init);
+program
+  .command("init")
+  .description("Initialize Hot Updater")
+  .addOption(
+    new Option(
+      "--provider <provider>",
+      "provider to use; skips the prompt",
+    ).choices(["cloudflare", "aws", "supabase", "firebase"]),
```
gronxb, line 89:
```suggestion
    ).choices(Object.keys(PROVIDER_LABELS)),
```

gronxb (changes requested):
How about this ?

gronxb:
It's a small but meaningful improvement. I also understand what you're trying to do, and I've thought about this before as well. As you said, to make it fully non-interactive, we need to change things one by one.

In practice, for Cloudflare, it already mostly works as long as `wrangler login` succeeds. I think we just need to somehow resolve AWS, Firebase, and Supabase.

gronxb (approved):

d4mr:
appreciate the fast turnaround on this.
thanks for building and maintaining hot-updater! its exactly what i was looking for :)

happy to take a crack at the AWS / Firebase / Supabase side you mentioned. before I do: do you have a preferred shape in mind? i was picturing a single --non-interactive (or --ci) flag threaded through each provider's runInit, where a missing .env.hotupdater value errors out instead of dropping into a prompt. I could also do the cloudflare fail-fast (error on missing env) as a smaller first step.

roadmap looks stale so not sure what's already in flight, so let me know where this fits / if someone's already on it.

gronxb:
It’s not on the roadmap because the roadmap only includes projects with major milestones. The roadmap is almost complete now, actually haha.

I think this is more of a usability improvement, and I do think it would make things much more convenient. I’ve already implemented this feature to some extent on the Cloudflare side. So even now, for Cloudflare, once the user has entered the input, that information is also stored in `.env.hotupdater`. Then, when `init` is run again, if `.env.hotupdater` is already populated, we reuse it as-is and avoid prompting for keyboard input.

I think that would be the first scope. After that, we may only need to consider things like `wrangler login`, which you were concerned about. Or feel free to suggest another approach, or open an issue so we can continue the discussion. I may not reply quickly, but I should be able to track it periodically.

PR #1069, opened 2026-06-25
feat(cli): support multiple ids in `bundle delete`

valehelle:
## What

`hot-updater bundle delete` can now delete more than one bundle at a time:

- `hot-updater bundle delete <id1> <id2> ...`

The single-id form (`bundle delete <id>`) is unchanged.

## Why

Cleanup scripts (e.g. tearing down a review/preview channel when a PR closes) end up doing `bundle list -c <channel> --json`, parsing ids, and looping `bundle delete <id>` one at a time — committing on every single delete. Multi-id delete lets them remove the whole set in a single pass with one commit.

## How

- `delete` now takes a variadic `<bundle-ids...>`.
- All targets are deleted, then a **single `commitBundle()`** runs — fewer management-index rewrites and CDN invalidations than committing per bundle.
- Each removal is verified via the existing `waitForDeletedBundle`; the command exits non-zero if any bundle survives.

## Tests

`bundle.spec.ts` covers multi-id delete with a single commit and the no-ids validation error; existing single-id tests use the new array signature.

Locally: `bundle.spec.ts` passes (20/20), `oxfmt`/`oxlint` clean on the changed files, and `tsc --noEmit` is clean.

## Note

An earlier revision also supported `bundle delete --channel <channel>`. Removed per review — whole-channel deletion risks deleting production bundles by mistake. To clear a channel, list it and pass the explicit ids to the multi-delete command.


gronxb (changes requested):
Thanks for the contribution! Would you mind removing just the channel deletion part?

Deleting bundles through channel deletion may seem convenient at first glance, but to me it introduces a security risk because it could result in deleting production bundles by mistake. That said, I think supporting the deletion of multiple bundles at once is perfectly reasonable.

For that reason, I think it’s better to intentionally favor safety over convenience. If someone really needs to delete all bundles in a channel, they can first list the channel’s bundles and then explicitly pass those bundle IDs to the multi-delete command. I think that workflow is the safer approach.

valehelle:
Thanks for the review — makes sense, agreed on favoring safety. I've **removed the channel-deletion path entirely** (the `--channel`/`--platform` options and the cursor-paginated enumeration) in d1347ad, and kept only multi-id `bundle delete <id...>`. To clear a channel, you now list it and pass the explicit ids, as you suggested.

Tests updated accordingly (dropped the channel-pagination + both-args validation cases; 20/20 passing), `oxfmt`/`oxlint`/`tsc` clean. PR description updated too.

gronxb (approved):

gronxb:
Released in v0.35.2 Thanks !

Issue #1074, opened 2026-07-03
iOS downloaded bundle works after update, onNotifyAppReady returns STABLE, but app loads old bundle after restart


hoang1204:
I am using HotUpdater in a React Native app. On iOS, the app can successfully check for an update, download the new bundle, and apply it when I use forceUpdate.

After the update, the new JS bundle works correctly. Also, onNotifyAppReady is called and returns STABLE.

However, when I kill the app completely and open it again, the app loads the old bundle instead of the newly downloaded bundle.

Environment

- Platform: iOS
- HotUpdater: 0.35.0
- React native: 0.77.3
- old architecture
- Update strategy: appVersion
- Testing with forceUpdate
- The new bundle is uploaded successfully
- The app downloads and applies the new bundle successfully
- onNotifyAppReady returns STABLE
- After restarting the app, it loads the old bundle again

onNotifyAppReady is called and returns:

{
  status: "STABLE"
}

So I do not think this is a normal crash recovery rollback.

HotUpdater.getBundleId() return old bundle id

This makes me unsure whether the active downloaded bundle is being tracked correctly on iOS.

--It's work correctly in Android

gronxb:
Could you check native setting in AppDelegate.swift ??

hoang1204:
I have configured the native iOS setup in AppDelegate according to the documentation.
```js
- (NSURL *)bundleURL
{
#if DEBUG
  return [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index"];
#else
  return [HotUpdater bundleURL];
#endif
}
```
However, the issue still happens on iOS:

The iOS bundle only updates when I set forceUpdate: true.
When forceUpdate is not enabled, the bundle is not updated.
Even when I enable forceUpdate, the app downloads and applies the new bundle, but after I close and reopen the app, it loads the old bundle again.
Then the update process repeats again.

Could you suggest what else I should check on the native iOS side?

gronxb:
Could you first run `hot-updater doctor` and check whether the native iOS setup is detected correctly?

https://hot-updater.dev/docs/guides/doctor

Also, could you describe your test scenario in more detail?

I’d like to understand whether this is a normal update scenario where the downloaded bundle is expected to be compatible but is not persisted after restarting the app, or whether you are testing with an intentionally incompatible bundle and the app is rolling back as part of the recovery flow.

If this bundle is expected to be compatible, could you also check `HotUpdater.getCrashHistory()`?

[...]

PR #1077, opened 2026-07-07
test(e2e): cover iOS compatible update cold restart

gronxb:
## Summary

- Adds a `compatible-update-cold-restart` Detox scenario for the issue #1074 repro shape: normal compatible production bundle install, stable reload, full app terminate, cold relaunch, then bundle identity checks.
- Converts the iOS v0.85 example startup path to use `SceneDelegate`, so the repro now exercises a scene-based cold launch instead of the previous AppDelegate-only window path.
- Extends `force-update-auto-reload` with the same cold-start identity checks so forced updates also prove persistence across a real app restart.
- Captures post-reload and cold-relaunch state snapshots, and asserts runtime bundle id, `STABLE` launch status, empty crash history, and active metadata.

## Context

Reproduction probe for https://github.com/gronxb/hot-updater/issues/1074#issuecomment-4899072366.

The target comment says the goal is a normal compatible update: iOS downloads/installs the bundle, `notifyAppReady` reports `STABLE`, crash history is empty, but after closing and reopening the app it loads the old bundle and repeats the update flow. The new scenario makes that exact cold-start boundary part of the default Detox suite, and the latest commit routes the iOS app through `SceneDelegate` to match the reported app shape more closely.

## Validation

- Red check before adding the normal scenario: `pnpm -w exec vitest run e2e/detox/scenario-contract.spec.ts -t "compatible-update-cold-restart"` failed with `Unknown Detox scenario: compatible-update-cold-restart`.
- Green targeted check: `pnpm -w exec vitest run e2e/detox/scenario-contract.spec.ts -t "compatible-update-cold-restart"`.
- Green full contract check: `pnpm -w exec vitest run e2e/detox/scenario-contract.spec.ts`.
- Green repo lint: `pnpm -w lint`.
- Local plist/diff checks after adding SceneDelegate: `plutil -lint examples/v0.85.0/ios/HotUpdaterExample/Info.plist`, `git diff --check`, and full scenario contract test passed.
- Local direct `xcodebuild` is blocked by stale local Pods privacy file paths in `examples/v0.85.0/ios/Pods.xcodeproj`; the authoritative native build/runtime validation is the `hot-updater-agent` standalone run below.

## Hot Updater Agent

- Previous `standalone-s3` iOS run on `ecc9fd3d` passed, but that run did not use `SceneDelegate`.
- Latest `standalone-s3` iOS run on `a8ccff08` used the SceneDelegate startup path. Job `job-20260707020312-wknnb5` reached and passed `compatible-update-cold-restart`, including post-reload and cold-relaunch bundle identity, `STABLE` launch report, empty crash history, and active metadata checks.
- The same latest full suite later failed in unrelated disabled-bundle rollback scenarios after the standalone storage/API path returned `connect ECONNREFUSED 127.0.0.1:19007`; final count was `13 passed, 2 failed, 15 total`.

### DIFF
--- a/e2e/detox/scenario-contract.spec.ts
+++ b/e2e/detox/scenario-contract.spec.ts
@@ -1545,7 +1545,8 @@ describe("Detox scenario contract", () => {
     const stages = await scenarioStages("force-update-auto-reload");
 
     // When: the Detox scenario is inspected.
-    // Then: it waits pending first, reloads, then verifies the stable launch.
+    // Then: it waits pending first, reloads, then proves the active bundle
+    // survives a cold relaunch.
     expect(stages).toEqual([
       "deploy force update bundle",
       "launch force update app",
@@ -1553,7 +1554,18 @@ describe("Detox scenario contract", () => {
       "wait force update metadata pending",
       "reload force update",
       "wait force update metadata stable",
+      "assert force update runtime bundle",
       "assert force update launch",
+      "assert force update crash history empty",
+      "capture force update post-reload state",
+      "assert force update metadata active",
+      "terminate force update app",
+      "cold relaunch force update app",
+      "assert force update cold runtime bundle",
+      "assert force update cold launch",
+      "assert force update cold crash history empty",
+      "capture force update cold state",
+      "assert force update cold metadata active",
     ]);
     expect(
       (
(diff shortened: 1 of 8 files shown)


gronxb:
Closing this repro probe. The follow-up request is being moved back to #1074.

PR #1081, opened 2026-07-09
fix(expo): disable minify when Hermes is enabled

### LINKED ISSUE #1080: Expo plugin: missing `--minify false` for Hermes — terser-minified bundle miscompiled by Hermes V1 (RN 0.85), OTA-only render crash
### Summary

The Expo build plugin (`@hot-updater/expo`) does not pass `--minify false` to `expo export:embed` when Hermes is enabled. The bare plugin already does this since #309, and the official React Native embed pipeline (`react-native-xcode.sh` / gradle) has always done it ("Hermes doesn't require JS minification", see also facebook/hermes#452).

As a result, OTA bundles are built from **terser-minified JS → hermesc**, a combination the embedded (store) build never produces. On RN 0.85 / Expo SDK 56 (**Hermes V1** toolchain, `hermes-compiler` npm package), hermesc's default optimizer **miscompiles certain terser output patterns**, producing deterministic, OTA-only runtime crashes that do not reproduce in the embedded binary built from the exact same source.

### Environment

- hot-updater 0.35.1 (`@hot-updater/expo`, `updateStrategy: "fingerprint"`, `compressStrategy: "tar.gz"`)
- Expo SDK 56 / React Native 0.85.3 (Hermes V1, New Architecture), `hermes-compiler` 250829098.0.10
- Reproduced on both iOS and Android (identical stack coordinates)

### Symptom

Entering a screen that renders remote HTML via `react-native-render-html` crashes with:

```
TypeError: Cannot read property 'children' of undefined
componentStack: at RenderTTree (react-native-render-html/src/RenderTTree.tsx:8) ...
```

- Embedded release build from the same commit: works.
- OTA bundle deployed from the same commit via `hot-updater deploy`: crashes, on both platforms.
- App boots fine; the crash is render-time in a specific subtree. Inputs were verified sane at runtime (the fetched HTML string reaches the component intact).

### Root-cause isolation (single-variable experiments)

Simulator experiment swapping only `main.jsbundle` inside the same release app (native binary, env, account, data all fixed):

| JS input to hermesc | hermesc flags | Result |
|---|---|---|
| `--minify false` (same as embedded pipeline) | default (`-O`) | ✅ OK |
| minified (current expo plugin behavior) | default (`-O`) | 💥 crash — identical coordinates to production OTA |
| minified (same file) | `-O0` | ✅ OK → optimizer-stage miscompile |
| minified, with a semantically-neutral `try/catch` added around the affected call | default (`-O`) | ✅ OK (heisenbug — shape-sensitive) |

Additional notes:

- Both hermesc binaries (Pods `hermes-engine` and npm `hermes-compiler`) are the same version — compiler binary difference ruled out.
- With this hermesc, `-O` is the default (`-O` and no flag produce byte-identical output; only `-O0` differs).
- The same pipeline on SDK 54 (previous Hermes generation) did not crash, which is consistent with a Hermes V1 optimizer regression being the trigger. The minified→hermesc input class is simply not exercised anywhere else in the RN ecosystem, since the official embed pipeline always disables minification for Hermes.

### Fix

Pass `--minify false` to `expo export:embed` when Hermes is enabled, exactly like the bare plugin does since #309. After patching this locally and redeploying, the crash is gone on both platforms (verified with the deployed artifacts). The resulting `.hbc` is even marginally smaller (minified input inflates the debug source-location tables).

PR: (linked below)

Unuuuuu:
Closes #1080

### What

Pass `--minify false` to `expo export:embed` when Hermes is enabled — same behavior as the bare plugin (#309) and the official React Native embed pipeline (`react-native-xcode.sh` / gradle; "Hermes doesn't require JS minification", facebook/hermes#452).

### Why

Without this flag, `expo export:embed --dev false` defaults to minify=true, so OTA bundles are built from **terser-minified JS → hermesc** — a combination the embedded build never produces. On RN 0.85 (Hermes V1), hermesc's default optimizer miscompiles certain terser output patterns, causing deterministic OTA-only runtime crashes (see #1080 for the full single-variable isolation matrix: unminified=OK / minified=crash / minified+`-O0`=OK).

### Notes

- Reuses the existing `isHermesEnabled` check (hoisted above the args array), mirroring the bare plugin's `String(!enableHermes)` style and comment.
- Verified in production: after applying this change and redeploying, the crash disappeared on both iOS and Android; the resulting `.hbc` is marginally smaller than the minified-input build.

### DIFF
--- a/.changeset/fancy-meteors-push.md
+++ b/.changeset/fancy-meteors-push.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/expo": patch
+---
+
+fix(expo): disable minify when Hermes is enabled
--- a/plugins/expo/src/expo.ts
+++ b/plugins/expo/src/expo.ts
@@ -53,6 +53,7 @@ const runBundle = async ({
   const bundleOutput = path.join(buildPath, `${filename}.bundle`);
   const entryFile = resolveMain(cwd);
   const bundleId = uuidv7();
+  const enableHermes = isHermesEnabled(cwd, platform);
 
   const args = [
     "expo",
@@ -65,6 +66,9 @@ const runBundle = async ({
     bundleOutput,
     "--dev",
     String(false),
+    // disable minify when enableHermes is true
+    "--minify",
+    String(!enableHermes),
     "--assets-dest",
     buildPath,
     ...(sourcemap ? ["--sourcemap-output", `${bundleOutput}.map`] : []),
@@ -86,7 +90,6 @@ const runBundle = async ({
     }
   }
 
-  const enableHermes = isHermesEnabled(cwd, platform);
   if (enableHermes) {
     const { hermesVersion } = await compileHermes({
       cwd,


gronxb (approved):
I can’t believe this wasn’t handled. Thanks for catching it! If Hermes is enabled, minification should be disabled.

gronxb:
Released in 0.35.3. Thanks !

Issue #960, opened 2026-04-25
Infinite loop crash


ehynds:
### Precheck

- [x] Infrastructure may have changed. Did you try `hot-updater doctor`? (https://hot-updater.dev/docs/guides/doctor)
- [x] Is your Node version v20 or higher?
- [x] Are you aware that the update will only work properly if you `hot-updater deploy` after building the native app?

### Name, Version of hot-updater

0.30.2

### Version of react-native

0.84.1

### Platforms

Android

### Architecture

New Architecture(Fabric)

### What provider are you seeing the problem on? (multiple)

S3 & Lambda

### Expo App

- [ ] I am using Expo
- [ ] Project is managed with Expo [CNG](https://docs.expo.dev/workflow/continuous-native-generation/)(android, ios directories are included in `.gitignore`)

### Bundler

metro

### What happened?

A small subset of my users are experiencing a `Maximum update depth exceeded` crash. Here's the stacktrace from Sentry:

<img width="845" height="621" alt="Image" src="https://github.com/user-attachments/assets/847c558d-8ec4-411e-8a3d-75464f62e28e" />

I had Claude do an investigation, and it came back with this:

> The crash is an infinite re-render loop in @hot-updater/react-native@0.30.2's store, triggered during the auto-update flow.
>
> The problem is in store.js — setState always creates a new state object reference and always calls emitChange(), even when the values haven't changed:
>
> ```
>  // store.js:22-38
>  const setState = newState => {
>     const nextState = { ...state, ...newState };  // always new object
>     // ...
>     state = nextState;
>     emitChange();  // always notifies listeners — no bailout
>  };
> ```
>
> Combined with `addListener("onProgress", setState)` on line 47, every native `onProgress` event — even if the progress value is identical — creates a new state object and forces React to schedule a synchronous re-render via `forceStoreRerender` → `enqueueConcurrentRenderForLane`.
>
> In your app (Fabric + bridgeless, confirmed by the event tags), native events can dispatch synchronously during a React render commit. When the native module fires rapid onProgress events during the update check/download, each one triggers a nested re-render. After ~50 nested updates, React throws "Maximum update depth exceeded."
>
>  The breadcrumbs confirm the sequence:
>  1. App starts → navigates to Routes
>  2. Auto-update check fires → GET .../check-update/... returns 200 at 15:37:18
>  3. Native module starts emitting onProgress events
>  4. User taps Franklin at 15:37:20
[...]

PR #1087, opened 2026-07-15
fix(react-native): coalesce progress store notifications

khethelogp:
### Bug

#960 / #966 fixed the crash for *duplicate* progress events by bailing out when the
snapshot is unchanged. However, during a real bundle download, native emits a rapid
burst of *distinct* events (progress/downloadedBytes advance every event). Each one
passes the equality check and synchronously notifies React — on Fabric/bridgeless
Android the burst exceeds React's nested update limit and the app crashes during the
download:

```
FATAL EXCEPTION: mqt_v_native
com.facebook.react.common.JavascriptException: Error: Maximum update depth exceeded.
enqueueConcurrentRenderForLane → emitChange → setState → applyProgressEvent → emit
```

Reproduced on v0.35.3 (which includes #966), RN 0.84.1, new architecture, Android
emulator (API 36), archive artifact ~9 MB.

### Fix

`emitChange` coalesces notifications on a short trailing timer (32ms):

- state updates remain fully synchronous — `getSnapshot()` always returns the latest
  snapshot, so subscribers never observe stale data when the flush runs
- render pressure drops from one React notification per native event to ≤ ~30/s
- the flush runs as a separate task, so React's nested-update limit cannot trip

### Tests

- new spec: a burst of 200 distinct progress events keeps `getSnapshot()` current
  immediately and produces exactly one subscriber notification after the flush
  (this scenario crashed before the change)
- existing store specs adapted to fake timers — required because notification is
  now deferred; all state (`getSnapshot()`) assertions are unchanged

Also verified in a production RN 0.84 new-arch app by packing this branch and
re-running the original crash scenario — the download completes and the UI
updates correctly.

### Evidence

[hot-updater-crash-log.txt](https://github.com/user-attachments/files/30064932/hot-updater-crash-log.txt)

| Bug (v0.35.3) | With this fix |
|---|---|
|  |  |


packages/react-native/src/store.ts
```diff
@@ -91,10 +91,20 @@ const createHotUpdaterStore = () => {
...
+  let isEmitScheduled = false;
+
   const emitChange = () => {
-    for (const listener of listeners) {
-      listener();
+    if (isEmitScheduled) {
+      return;
     }
+    isEmitScheduled = true;
+    setTimeout(() => {
```
gronxb:
```suggestion
    requestAnimationFrame(() => {
```

Could you test this code ?

gronxb (commented):
Thanks for your contribution ! Please check one

packages/react-native/src/store.ts
khethelogp: You beat me to it 😄, that's a good suggestion, was going to make a fix in the afternoon

gronxb: haha. Thanks for the report. Reproduced and solved it.

gronxb (approved):

gronxb:
Released in v0.35.4 Thanks you so much !

PR #1095, opened 2026-07-25
fix(android): preserve brotli decoder when Android minification is actived

valeriobelli:
## Summary

When Android minification (R8) is enabled in release builds, R8 strips the static initializer of the bundled Brotli decoder's embedded dictionary (DictionaryData). This causes OTA `.tar.br` bundle extraction to fail at runtime with:

```
java.io.IOException: Brotli stream decoding failed
Caused by: brotli dictionary is not set
```

## Changes

Add a ProGuard/R8 -keep rule for com.hotupdater.vendor.brotli.** in proguard-rules.pro, preventing R8 from removing the Brotli decoder and its static dictionary so OTA bundle decompression continues to work in minified release builds.

## Impact

- Fixes OTA updates failing to apply on minified Android release builds.
- No effect on debug or non-minified builds.

### DIFF
--- a/.changeset/quick-meals-sit.md
+++ b/.changeset/quick-meals-sit.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+Preserve brotli decoder when Android minification is actived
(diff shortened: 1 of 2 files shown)


gronxb (approved):
Oops !

valeriobelli:
@gronxb supah fast merge, thanks! 

This can be also fixed downstream by hot-updater consumers, but shipping it by Hot Updater itself it's better.

PR #1108, opened 2026-07-29
chore: replace semver with verkit

gronxb:
## Summary

- Replace workspace-owned direct `semver` and `@types/semver` dependencies across 8 packages with exactly pinned `verkit@0.3.2`.
- Migrate runtime calls to Verkit named APIs while retaining `rangesIntersect` for range-to-range patch compatibility.
- Update inlined dependencies, the generated Supabase import-map expectation, the lockfile, and the existing patch changeset.
- Add a deployment regression test that prevents a prerelease target from using a stable wildcard range as an automatic patch base.

## Why 0.3.2

Verkit 0.3.1 incorrectly reported that `1.2.3-a` intersects `1.x`, which could select an incompatible automatic patch base. The upstream fix was merged in [sxzz/verkit#2](https://github.com/sxzz/verkit/pull/2) and released as 0.3.2.

The regression test was verified to:

- fail on `verkit@0.3.1` because `createBundleDiff` was called for the incompatible base
- pass on exactly pinned `verkit@0.3.2`

## Scope

Transitive `semver` copies required by third-party dependencies remain in `pnpm-lock.yaml`; this PR removes all direct workspace declarations and source imports.

All dependency and inlined-dependency declarations use the exact version `0.3.2`, without a caret or other range.

## Validation

- `pnpm -w build`
- `pnpm -w test:type` (33 projects)
- `pnpm -w lint`
- `pnpm -w test` (130 files, 1,886 tests)
- `pnpm -w test:integration` (15 files, 1,454 tests)
- `pnpm changeset status`
- Targeted deploy regression: 31 tests passed on 0.3.2
- Test integrity: no tests were deleted, skipped, or weakened; one behavioral regression test was added and the generated Supabase import-map expectation was updated to match the migration


gronxb:
@codex review

gameroman (approved):

gronxb:
@codex review

PR #1116, opened 2026-08-05
feat: platform-scoped fingerprint.extraSources (ios / android)

### LINKED ISSUE #1112: Platform-scoped fingerprint.extraSources (ios / android / shared)
## Problem

`fingerprint.extraSources` is a flat `string[]` and gets applied to **both** platform fingerprints.

That hurts brownfield / replace-in-place hosts where iOS and Android native inputs are different trees. An iOS-only host change (e.g. Info.plist embed hash) currently moves the Android fingerprint too, and the other way around — which forces unrelated native rebuilds and blocks OTA on the platform that didn't change.

## Proposal

Allow an object form in addition to `string[]`:

```ts
fingerprint: {
  extraSources: {
    shared?: string[]
    ios?: string[]
    android?: string[]
  }
}
```

When computing a platform fingerprint, use `shared + platform`. Keep today's `string[]` behavior as “apply to both”.

Also: the default config shouldn't set `extraSources: []` if that deep-merge can overwrite a user-supplied object with an empty array.

## Context

We've been carrying a tiny local patch for this in a brownfield Expo host (separate iOS/Android native release lines). Happy to open a PR if this direction makes sense.

glekner:
Closes #1112 — implements the shape you suggested in the issue.

## Summary

`fingerprint.extraSources` now accepts an object in addition to `string[]`:

```ts
extraSources: ["resources/**"]            // shared, both platforms (unchanged)
extraSources: { ios: ["ios/.env"] }       // iOS fingerprint only
```

When a platform fingerprint is computed, only the entries for that platform are
processed, so an iOS-only native input no longer moves the Android fingerprint
and vice versa. This matters for brownfield / replace-in-place hosts where iOS
and Android native inputs are different trees and ship on separate native
release lines — cross-platform contamination there forces unrelated native
rebuilds and blocks OTA on the platform that didn't change.

Fully backward compatible: the array form behaves exactly as before.

## Changes

- `plugin-core`: new exported `FingerprintExtraSources` type (`string[] | { ios?, android? }`), used by `ConfigInput["fingerprint"]["extraSources"]`.
- `hot-updater`: `resolveExtraSources(extraSources, platform)` in `processExtraSources.ts`; `getOtaFingerprintOptions` resolves per platform before processing.
- `cli-tools`: the default config no longer sets `fingerprint.extraSources: []`. `merge` (es-toolkit) would otherwise let that default array clobber a user-supplied object — verified: `merge({ extraSources: [] }, { extraSources: { ios: [...] } })` yields `[]`.
- docs: object form documented in the fingerprint strategy guide.
- changeset: patch for `hot-updater`, `@hot-updater/cli-tools`, `@hot-updater/plugin-core`.

## Test plan

- New `resolveExtraSources` unit tests (array shared / object scoped / missing platform / undefined).
- New `getOtaFingerprintOptions` tests asserting the processed `extraSources` per platform, including that a platform with no entries gets none.
- New `loadConfig` test proving the object form survives the config deep merge.
- `pnpm -w test` (157 files / 2081 tests), `pnpm -w lint`, `pnpm -w test:type` all pass.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/platform-scoped-fingerprint-extra-sources.md
+++ b/.changeset/platform-scoped-fingerprint-extra-sources.md
@@ -1,0 +1,16 @@
+---
+"@hot-updater/plugin-core": patch
+"@hot-updater/cli-tools": patch
+"hot-updater": patch
+---
+
+feat: support platform-scoped `fingerprint.extraSources`
+
+`fingerprint.extraSources` now accepts `{ ios?: string[], android?: string[] }`
+in addition to `string[]`. An array keeps the existing behavior (shared by both
+platforms); the object form only feeds the fingerprint of the platform it is
+scoped to, so an iOS-only native input no longer moves the Android fingerprint
+(and vice versa).
+
+The default config no longer sets `extraSources: []`, which the config deep
+merge would otherwise use to clobber a user-supplied object.
--- a/docs/content/docs/guides/update-strategies/fingerprint.mdx
+++ b/docs/content/docs/guides/update-strategies/fingerprint.mdx
@@ -47,6 +47,24 @@ export default defineConfig({
 });
 ```
 
+An array is shared: every entry is hashed into both the iOS and the Android
+fingerprint. When the native inputs differ per platform, use the object form so
+that an iOS-only change does not move the Android fingerprint (and vice versa):
+
+```typescript title="hot-updater.config.ts"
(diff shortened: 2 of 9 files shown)


gronxb (approved):
Looks good ! Thanks

Issue #1123, opened 2026-08-05
deploy hangs for ~16 minutes after "Upload Complete" on a large project.pbxproj


LonelyCpp:
We hit this on our CI OTA lane and it took a while to pin down, so writing it up in
case it's useful to someone else. Opening a PR alongside.

hot-updater 0.25.10, RN 0.83.2, bare, S3 storage + standalone repository, iOS. I also
read through 0.35.10 and this code path is unchanged there.

Our `hot-updater deploy --platform ios` step takes 21 minutes. Bundling accounts for
about 4 of those, so we assumed the upload was the problem. It isn't:

- bundle is a 30.7MB tar.gz, uploaded in 0.85s across 7 parts (31-61 MB/s)
- `POST /api/bundles` takes 121ms
- the gap between `✅ Upload Complete` and the database write is **995 seconds**

We had a 5s heartbeat timer running inside the deploy process while debugging. It
fired every 5s right up to the upload, then didn't fire again for 995s. An unref'd
interval that can't fire means the event loop is blocked, not waiting on IO. RSS went
from 189MB to 1.5GB across the same window.

The only thing between those two log lines is `getNativeAppVersion(platform)` at the
top of the "Updating Database" task, which for iOS is `getIOSVersionFromXcodeProject()`
-> `XcodeProject.open(pbxproj).toJSON()`, and that's synchronous.

Our `project.pbxproj` is 12.5MB. It has a lot of junk in it that shouldn't be there and
we're cleaning that up separately, but the parse cost is worth reporting on its own.
Timing just that call:

| @bacons/xcode | XcodeProject.open() | peak RSS |
|---|---|---|
| 1.0.0-alpha.24 (bundled in hot-updater) | 286s | 1796MB |
| 1.0.0-alpha.33 | 0.2s | 285MB |

Identical output from both, 48,670 objects and the same version string.

alpha.24 was published 2024-12-09. The difference is EvanBacon/xcode#37, which replaced
the Chevrotain parser with a single-pass one. Their benchmarks say 42x; on a file this
size it's far more than that, so the old parser looks like it degrades superlinearly.

Two things fix this and they're independent, so I've put both in a PR as
separate commits:

https://github.com/gronxb/hot-updater/pull/1124

1. Try `info-plist` before `xcodeproj` in `getNativeAppVersion`. Reading Info.plist is a
   small file read, and `CFBundleShortVersionString` is arguably the better source
[...]

Issue #84, opened 2025-02-05
[cli] CLI picks up version from MARKETING_VERSION but ideally should pick CFBundleShortVersionString


regalstreak:
Noticed this thing where while running `yarn hot-updater deploy -i`, the default version shows up as the `MARKETING_VERSION` in project.pbxproj, ideal DX should pick it up from the Info.plist as that's picked up during updates if I'm right?

Don't know why we're not updating the `MARKETING_VERSION` though 🥲

![Image](https://github.com/user-attachments/assets/41d47ba6-1e3e-4890-ad00-b8e7391341f7)

gronxb:
Great point! But I don’t have time right now, and since it’s a low priority, I’ll check it later when I have some free time.

It looks like I need to update the following code:
https://github.com/gronxb/hot-updater/blob/b0f645155ca1c8ae15248ca857ac9edd683ebae4/packages/hot-updater/src/utils/getDefaultTargetAppVersion.ts#L32-L63

PR #1124, opened 2026-08-05
fix: prefer Info.plist over project.pbxproj when resolving the iOS app version

LonelyCpp:
`getNativeAppVersion("ios")` runs the `xcodeproj` parser first, so every `deploy`
synchronously parses project.pbxproj just to fill in `metadata.app_version`. On our
project that's a 12.5MB pbxproj and it blocks the event loop for ~16 minutes on CI,
after the bundle has already finished uploading. Full trace in #1123.

Two commits, independent of each other, drop either if you disagree.

**Info.plist first.** Flips the order to `["info-plist", "xcodeproj"]`. Reading
Info.plist is a small file read, and `CFBundleShortVersionString` is closer to what the
built app actually reports than `MARKETING_VERSION`, which is what #84 was asking for on
correctness grounds. xcodeproj stays as the fallback, so anything that resolves today
still resolves.

**@bacons/xcode 1.0.0-alpha.24 -> 1.0.0-alpha.33.** alpha.24 is from December 2024;
alpha.31 picked up EvanBacon/xcode#37, the single-pass parser rewrite. This matters even
with the reorder, because everything else still goes through the xcodeproj path: the
`app-version` command, `getDefaultTargetAppVersion`, and the fallback above for projects
whose Info.plist uses `$(MARKETING_VERSION)`. The only API this repo uses is
`XcodeProject.open().toJSON()` in `getIOSVersion`, and it's unchanged across those
versions. I checked by building the CLI from this branch, reverting just the parser order
so it took the xcodeproj path, and running `app-version` against our 12.5MB pbxproj:
0.37s, same output as before. Same call on alpha.24 takes 286s.

On the tests: the existing xcodeproj test is now the fallback path, so I renamed it and
made it mock Info.plist away explicitly rather than relying on it happening to fail. The
second test now asserts Info.plist wins and that `XcodeProject.open` is never called.
That last assertion is the one actually protecting the fix, and it fails on main.

`pnpm -w lint` is clean, `pnpm -w build` succeeds, and `tsc --noEmit` in
packages/hot-updater is clean after a build. `pnpm -w test` has 72 failing files on main
from unbuilt workspace packages; the counts are identical with and without these changes
(72 failed / 85 passed, 46 / 596 / 1).

### DIFF
--- a/.changeset/hip-moons-invite.md
+++ b/.changeset/hip-moons-invite.md
@@ -1,0 +1,15 @@
+---
+"hot-updater": patch
+---
+
+fix: read the iOS app version from Info.plist before project.pbxproj
+
+`getNativeAppVersion("ios")` tried the `xcodeproj` parser first and only fell back
+to `info-plist`. Parsing project.pbxproj is synchronous, so on a large project it
+blocks the event loop for the whole parse, and `deploy` does this after the bundle
+has already been uploaded, just to fill in `metadata.app_version`. On a 12.5MB
+pbxproj that was ~5 minutes locally and ~16 minutes on CI.
+
+Info.plist is read first now. `CFBundleShortVersionString` is also closer to what
+the built app actually reports than `MARKETING_VERSION` (#84). The xcodeproj parser
+is still there as a fallback when Info.plist has no version.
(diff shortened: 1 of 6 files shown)


gronxb (approved):

gronxb:
Thanks for your contribution  ..!

PR #1127, opened 2026-08-06
fix(ios): fallback unresolved app version settings

gronxb:
## Summary

- treat unresolved Xcode build-setting references in `CFBundleShortVersionString` as unavailable
- fall back to the upgraded `project.pbxproj` parser merged in #1124
- add a regression test and patch changeset

Follow-up to #1124 for the `$(MARKETING_VERSION)` case described in #1123.

## Root cause

After #1124, Info.plist is checked first. When it contains `$(MARKETING_VERSION)`, the plist parser returns that placeholder as a non-empty string, so the parser chain stops before the xcodeproj fallback.

The fix rejects unresolved `$(` values at the Info.plist boundary. The existing xcodeproj parser then resolves `MARKETING_VERSION`; because #1124 also upgraded `@bacons/xcode`, this fallback uses the fast parser rather than the old multi-minute path.

## RED / GREEN

- RED: the new regression test returned `$(MARKETING_VERSION)` instead of `1.0`
- GREEN: the same test returns `1.0` and verifies `XcodeProject.open()` was called

## Validation

- `pnpm nx build hot-updater`
- `pnpm exec vitest run packages/hot-updater` (29 files, 291 tests)
- `pnpm --filter hot-updater test:type`
- `pnpm -w lint` (0 warnings, 0 errors)
- built CLI from `examples/v0.85.0`: `app-version --json` returned `{ "android": "1.0", "ios": "1.0" }`

### DIFF
--- a/.changeset/warm-dodos-wave.md
+++ b/.changeset/warm-dodos-wave.md
@@ -1,0 +1,5 @@
+---
+"hot-updater": patch
+---
+
+fix: fallback to project.pbxproj when Info.plist contains an unresolved build setting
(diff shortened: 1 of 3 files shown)


gronxb:
@codex review

PR #1143, opened 2026-08-12
fix(server): treat unknown built-in baseline bundle ids as nil baseline

### LINKED ISSUE #1015: AWS Lambda should treat unknown built-in baseline bundle ids as nil instead of returning 503
# AWS Lambda should treat unknown built-in baseline bundle ids as nil instead of returning 503

## Summary

The AWS check-update runtime can fail before returning an available OTA when the client reports a generated built-in baseline bundle id that does not exist in the Hot Updater database.

This happened on iOS with a TestFlight build that had not installed any OTA bundle yet.

Instead of behaving like a fresh built-in app baseline, the request included a generated bundle id:

```text
019e37fe-bd30-7000-8000-000000000000
```

The AWS runtime had no database record for that id, so the request failed before it could return the latest compatible OTA.

---

## Observed Request Shape

```text
/ios/1.2.0/production/019e37fe-bd30-7000-8000-000000000000/019e37fe-bd30-7000-8000-000000000000/452
```

---

## Actual Behavior

The Lambda check-update request fails instead of returning the available OTA.

In our case, an enabled iOS OTA existed:

```text
019e3f00-90cc-70c9-bb2b-b7cf2acecf60
```

but the request died while resolving the unknown current baseline bundle id, before that OTA could be returned.

---

## Expected Behavior

A fresh native app bundle with no installed OTA should be treated as the built-in nil baseline:

```text
00000000-0000-0000-0000-000000000000
```

The healthy request shape should be:

```text
/ios/1.2.0/production/00000000-0000-0000-0000-000000000000/00000000-0000-0000-0000-000000000000/452
```

That request should return:

```text
UPDATE
```

with the latest compatible OTA bundle id.

---

## Why This Matters

A generated built-in baseline id is not a deployed OTA bundle id.

If the runtime requires every incoming current/min bundle id to exist in the OTA database, fresh native installs can fail to receive their first update.

This breaks first-launch OTA delivery for store/TestFlight builds that report a generated baseline id instead of the nil built-in baseline.

---

## Suggested Runtime Behavior

The AWS runtime should handle this defensively:

- If the incoming current/min bundle id is the nil UUID, treat it as a built-in baseline.
- If the incoming current/min bundle id does not exist in the database and appears to be a native built-in baseline, do not throw or return 503.
- Treat it as equivalent to the nil built-in baseline and return the latest compatible OTA as a full update when available.

Pseudo behavior:

```ts
if (requestedBundleId === NIL_UUID) {
  return resolveFromBuiltInBaseline();
}

if (!database.hasBundle(requestedBundleId)) {
  return resolveFromBuiltInBaseline();
}
```

---

## Related Native-Side Fix

The cleaner native contract is for a fresh built-in app bundle to report:

```text
00000000-0000-0000-0000-000000000000
```

instead of generating a timestamp-shaped id when no OTA bundle is installed.

That keeps the runtime request unambiguous and prevents the AWS checker from trying to resolve a non-existent OTA bundle id.

---

## Reproduction Context

- Platform: iOS
- Distribution: TestFlight
- App version: `1.2.0`
- Build: `24`
- Update strategy: `appVersion`
- Channel: `production`
- Enabled OTA bundle existed, but update check failed before returning it.

---

## Expected Outcome

The AWS runtime should not fail on unknown built-in baseline ids.

It should gracefully treat them as the nil baseline and return the latest compatible OTA update when one exists.

giaBaoJS:
## Root cause

Fixes #1015

A fresh install that has never applied an OTA update reports its **build-time generated built-in baseline bundle id** (from `generateMinBundleId()`, e.g. `019e37fe-bd30-7000-8000-000000000000`) as its current bundle id. That id is derived from the build timestamp with all UUIDv7 random bits zeroed, so it never exists as a deployed bundle record.

The update-check path still treated it like a regular installed bundle id:

- In the shared `getUpdateInfo` core (`@hot-updater/js`, used by the AWS Lambda runtime through `resolveUpdateInfoFromBundles`, and also by the Cloudflare/Firebase providers and the server DB adapters), an unknown baseline id could enter **rollback resolution** (`ROLLBACK` to an older bundle, or the forced nil-bundle `ROLLBACK`) instead of behaving like the nil baseline `00000000-0000-0000-0000-000000000000`.
- In `@hot-updater/server`'s `getAppUpdateInfo`, the unknown baseline id was **resolved against the database** (`getBundleById`) when no request-scoped seeds are available. On the AWS S3 database plugin that lookup cache-misses and falls back to a full bundle re-scan inside Lambda@Edge — the "request died while resolving the unknown current baseline bundle id" failure reported in the issue, surfacing as a CloudFront 503 before the available OTA could be returned.

## Fix

Applied in the shared core so every provider that funnels through it benefits (AWS, Cloudflare, Firebase, server DB adapters):

- `packages/core/src/uuid.ts`: add `isBuiltInBaselineBundleId()` — detects the `generateMinBundleId()` shape (`xxxxxxxx-xxxx-7000-8000-000000000000`).
- `plugins/js/src/getUpdateInfo.ts`: when the reported bundle id is a built-in baseline id **and has no bundle record**, evaluate the request with nil-baseline semantics. A fresh install therefore receives the latest compatible enabled OTA as `UPDATE` (never `ROLLBACK`), and gets `null`/`UP_TO_DATE` when nothing is available. Unknown ids that are not baseline-shaped keep the existing rollback semantics (deleted-bundle recovery is unchanged).
- `packages/server/src/db/pluginCore.ts`:
  - `getAppUpdateInfo` no longer resolves a built-in baseline id against the database — the record cannot exist, and on AWS the lookup is a full S3 re-scan inside Lambda@Edge.
  - `findUpdateInfoByScanning` (generic scanning fallback) applies the same nil-baseline semantics.

Request from the issue (`/ios/1.2.0/production/019e37fe-…/019e37fe-…/452` with enabled OTA `019e3f00-…`) now resolves to that OTA with `status: "UPDATE"`.

Note: the SQL-side implementations (Supabase RPC / self-hosted Postgres functions) evaluate update checks in SQL and are untouched here; they can be aligned in a follow-up if desired.

## Tests

- `plugins/js/src/getUpdateInfo.spec.ts` — 8 new tests, including the exact ids from the issue: unknown baseline id + newer enabled OTA → `UPDATE`; unknown baseline id with only stale bundles → treated as nil baseline (no bogus rollback); unknown **non**-baseline ids still roll back; baseline-shaped id that actually exists in the DB is left untouched; fingerprint-strategy variants.
- `packages/server/src/db/pluginCore.spec.ts` — verifies `getAppUpdateInfo` never issues a `getBundleById` for a built-in baseline id, and that the scanning fallback treats it as the nil baseline.
- `plugins/aws/src/s3Database.spec.ts` — reproduces the issue scenario against the S3 database plugin (unknown baseline id, enabled newer OTA for the same app version → `UPDATE`, with no S3 object listing).
- `plugins/plugin-core/src/generateMinBundleId.spec.ts` — asserts generated min-bundle ids always match `isBuiltInBaselineBundleId()` so the pattern cannot drift.

## Test evidence

- New js-core and pluginCore tests fail without the fix (3 + 2 failures) and pass with it
- `pnpm build` — all packages/plugins build
- `pnpm test` — full unit suite passes
- `pnpm test:type` — passes for all projects
- `pnpm lint` — oxfmt/oxlint clean on touched files

### DIFF
--- a/.changeset/unknown-baseline-bundle-id.md
+++ b/.changeset/unknown-baseline-bundle-id.md
@@ -1,0 +1,7 @@
+---
+"@hot-updater/core": patch
+"@hot-updater/js": patch
+"@hot-updater/server": patch
+---
+
+Treat unknown build-time generated baseline bundle ids as the nil baseline during update checks so a fresh install still receives the latest compatible OTA instead of failing or rolling back.
--- a/packages/core/src/uuid.ts
+++ b/packages/core/src/uuid.ts
@@ -1,1 +1,19 @@
 export const NIL_UUID = "00000000-0000-0000-0000-000000000000";
+
+/**
+ * Suffix shared by every build-time generated baseline bundle id
+ * (`generateMinBundleId()`): a UUIDv7 whose random bits are all zero.
+ *
+ * A fresh install that has never applied an OTA update reports this id as its
+ * current bundle id even though it is never stored as a deployed bundle.
+ */
+const BUILT_IN_BASELINE_BUNDLE_ID_SUFFIX = "-7000-8000-000000000000";
+
+/**
+ * Whether the given bundle id is a build-time generated built-in baseline id.
+ *
+ * Such ids identify the JS bundle embedded in the native app binary, not a
+ * deployed OTA bundle, so they never have a database record.
+ */
+export const isBuiltInBaselineBundleId = (bundleId: string): boolean =>
+  bundleId.endsWith(BUILT_IN_BASELINE_BUNDLE_ID_SUFFIX);
--- a/packages/server/src/db/pluginCore.spec.ts
+++ b/packages/server/src/db/pluginCore.spec.ts
@@ -422,6 +422,144 @@ describe("createPluginDatabaseCore", () => {
     expect(getBundleById).toHaveBeenCalledWith(targetBundle.id, undefined);
   });
 
+  it("skips database resolution for unknown built-in baseline bundle ids", async () => {
+    const baselineBundleId = "019e37fe-bd30-7000-8000-000000000000";
+    const targetBundle = {
+      ...baseBundle,
+      id: "019e3f00-90cc-70c9-bb2b-b7cf2acecf60",
+      fileHash: "hash-2",
+      manifestStorageUri: "s3://bucket/target/manifest.json",
+      manifestFileHash: "sig:target-manifest",
+      assetBaseStorageUri: "s3://bucket/target/files",
+    };
+    const manifests = new Map([
+      [
+        targetBundle.manifestStorageUri,
+        JSON.stringify({
+          bundleId: targetBundle.id,
+          assets: {
+            "index.ios.bundle": {
+              fileHash: "target-bundle-hash",
+            },
+          },
+        }),
+      ],
+    ]);
+    const getBundleById = vi.fn<DatabasePlugin<TestContext>["getBundleById"]>(
+      async (bundleId) => (bundleId === targetBundle.id ? targetBundle : null),
+    );
+    const getUpdateInfo = vi.fn<
+      NonNullable<DatabasePlugin<TestContext>["getUpdateInfo"]>
+    >(async () => ({
+      fileHash: targetBundle.fileHash,
+      id: targetBundle.id,
+      message: targetBundle.message,
+      shouldForceUpdate: targetBundle.shouldForceUpdate,
+      status: "UPDATE",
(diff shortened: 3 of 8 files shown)


gronxb:
Sorry for not taking care of this PR. I'm currently preparing for the 1.0 release, and a lot of things are changing internally. I'll close this for now, but if this turns out to be a critical issue, I'll reopen it.

PR #1146, opened 2026-08-13
feat(storage): reduce S3 scans and add reference-aware pruning

### LINKED ISSUE #1131: AWS console performance
Coming back from #917 we've now upgraded to 0.35.11, haven't been using console that much but noticed it is quite slow now. Our state of things:

- bucket is shared across 3 envs (development, staging, production)
- bucket has now accumulated ~38.4GB, 1.84K total objects or 84 pages of bundles in console (84 * 20 = ~1 680 unique bundles)

Console load itself takes about a minute:
- network tab timing show `Waiting: 60840.7ms`
- result of `time pnpm hot-updater bundles list`: `pnpm hot-updater bundle list  9.06s user 5.38s system 22% cpu 1:02.83 total`

I'm filing this more as a question as I'm not quite sure if you can push it further with S3 "database" being just JSON files
It's now easier to cleanup bundles so I'll probably do that, is that what you'd recommend for now?

gronxb:
## Root cause

S3 management discovery treated every root-level legacy UUIDv7 bundle artifact directory as a possible metadata prefix. With about 1,680 legacy bundle directories, one unfiltered management query produced about 1,680 avoidable recursive `ListObjectsV2` calls. Repeated Console refetches then multiplied that work and saturated the AWS SDK socket pool.

The shared `assets/` directory was not the bundle-count-dependent cause: it adds one root traversal. The fan-out came from legacy `<bundle-id>/...` directories.

After this change, management discovery performs root pagination, channel/platform traversal, and canonical `update.json` reads. It no longer performs one recursive LIST per artifact directory. Root pages and metadata partitions still scale; this PR does not claim mathematically constant S3 metadata queries.

## What changed

- skip only legacy root UUIDv7 artifact prefixes during unfiltered S3 metadata discovery
- preserve valid `assets` and `bundles` channel names while avoiding traversal below their non-platform storage children
- cover the reported scale with a paginated 1,680-prefix regression test
- derive channels from canonical `update.json` keys without downloading every manifest
- store new archives, manifests, and patches below `bundles/<bundle-id>/...`
- continue reading and cleaning existing root `<bundle-id>/...` storage URIs without migration
- keep shared content-addressed assets below `assets/sha256/...`
- reject new UUIDv7 S3 channel names because they are indistinguishable from legacy artifact roots during unfiltered discovery
- resolve CLI and Console multi-delete targets from one management snapshot and commit once per batch
- tolerate duplicate and already-absent IDs in Console batch deletion
- add exact target app version filters to `bundle list` and the Console; an exact channel/platform/version scope reads only that metadata partition
- use one shared physical-name rule for deploy, promote, server artifact resolution, diff generation, and prune

## Storage pruning

AWS S3 storage supports paginated object listing and exact raw-key batch deletion:

```bash
# Explicit dry run; protects objects newer than 24h by default\npnpm hot-updater storage prune --dry-run

# Delete eligible objects after reviewing the dry run
pnpm hot-updater storage prune --min-age 7d --yes
```

The command finds:

- orphaned legacy `<bundle-id>/...` objects
- orphaned `bundles/<bundle-id>/...` objects
- unreferenced `assets/sha256/...` objects

Individual bundle deletion intentionally leaves shared content-addressed assets for prune. Console wording now distinguishes metadata deletion, best-effort per-bundle artifact cleanup, and shared-asset reclamation.

### Safety contract

- dry-run by default, with an explicit `--dry-run` flag for automation
- 24h minimum age by default; young and undated objects are protected
- loads every database page and aborts on missing or repeated cursors
- validates live manifest `bundleId` and SHA-256 asset hashes before listing or deleting
- aborts when a live content-addressed asset base uses a different protocol from the active storage plugin
- protects exact archive, manifest, and patch references plus legacy asset prefixes
- narrows legacy UUID matching to known artifact shapes so UUID-shaped metadata is not pruned
- deletes raw relative S3 keys, preserving `#`, `?`, `%`, spaces, and Unicode exactly
- reloads references before deletion, but does not claim atomic online GC

`storage prune --yes` is an exclusive maintenance operation: stop deploy and promote while it runs. The current database must own every object under the configured storage prefix. If databases or environments are separate, configure a separate storage `basePath` for each before pruning.

## Compatibility decisions

Kept because they represent real published layouts:

- legacy root `<bundle-id>/...` reads and cleanup
- legacy per-bundle `files/` assets
- optional storage-management capabilities for non-S3 providers

Removed because they were unsafe guesses rather than valid compatibility contracts:

- guessing a storage root when an upload URI does not contain the bundle ID
- treating arbitrary `main.<platform>.bundle` files as Brotli objects when runtime resolution only supports `index.<platform>.bundle`
- comparing CDN and S3 URI strings as if they proved object identity

## Long-term boundary

Future artifacts are grouped below one `bundles/` root, so artifact growth no longer increases root prefix cardinality. Unfiltered blob metadata management still reads canonical metadata partitions; a DynamoDB/index migration remains a separate long-term scalability option rather than a prerequisite for this fix.

Closes #1131

## Verification

- `pnpm -w lint`
- `pnpm -w test:type` — 34 projects
- `pnpm -w build` — 26 projects
- `pnpm -w test` — 162 files, 2,125 tests
- `pnpm -w test:integration` — 15 files, 1,454 tests
- built CLI help verified for `storage prune --dry-run`, its conflict with `--yes`, and `bundle list --target-app-version`

### DIFF
--- a/.changeset/fix-s3-management-listing.md
+++ b/.changeset/fix-s3-management-listing.md
@@ -1,0 +1,8 @@
+---
+"@hot-updater/aws": patch
+"@hot-updater/plugin-core": patch
+"@hot-updater/console": patch
+"hot-updater": patch
+---
+
+Reduce S3 management query work by excluding bundle artifact prefixes, deriving channels from canonical manifest keys, and batching multi-bundle deletion scans and commits.
--- a/.changeset/fix-s3-management-listing.md
+++ b/.changeset/fix-s3-management-listing.md
@@ -7,4 +7,4 @@
 "hot-updater": minor
 ---
 
-Reduce S3 management query work by excluding storage artifact prefixes, deriving channels from canonical manifest keys, and batching multi-bundle deletion scans and commits. Store new bundle artifacts below `bundles/<bundle-id>`, while preserving reads and cleanup for the legacy root layout. Add `hot-updater storage prune` to find orphaned bundle objects and unreferenced shared assets, with dry-run and minimum-age safeguards.
+Reduce S3 management query work by skipping legacy UUIDv7 artifact traversal, deriving channels from canonical manifest keys, and batching multi-bundle deletion scans and commits. Store new bundle artifacts below `bundles/<bundle-id>` while preserving legacy reads, and add exact target app version filters to the CLI and Console. Add an exclusive-maintenance `hot-updater storage prune` command for orphaned bundle objects and unreferenced shared assets, with dry-run, minimum-age, and fail-closed reference validation safeguards.
--- a/.changeset/fix-s3-management-listing.md
+++ b/.changeset/fix-s3-management-listing.md
@@ -7,4 +7,4 @@
 "hot-updater": minor
 ---
 
-Reduce S3 management query work by skipping legacy UUIDv7 artifact traversal, deriving channels from canonical manifest keys, and batching multi-bundle deletion scans and commits. Store new bundle artifacts below `bundles/<bundle-id>` while preserving legacy reads, and add exact target app version filters to the CLI and Console. Add an exclusive-maintenance `hot-updater storage prune` command for orphaned bundle objects and unreferenced shared assets, with dry-run, minimum-age, and fail-closed reference validation safeguards.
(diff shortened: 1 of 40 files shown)


gronxb:
@codex review

Issue #1145, opened 2026-08-13
Supabase: update-check fans out one `createSignedUrl` request per changed asset, exhausting the project's Postgres connection slots /


dadafly1244:
# Supabase: update-check fans out one `createSignedUrl` request per changed asset, exhausting the project's Postgres connection slots

## Summary

On a Supabase-backed setup, a single update check that returns an update issues **one `createSignedUrl` call to the Storage API per changed asset**, all fired concurrently. Because `hot-updater` forces the storage bucket to be **private**, every one of those calls goes through `storage-api`, which holds its own Postgres connection pool.

Under normal post-deploy traffic this ramps the Storage API pool up and it never shrinks back. On my project, **Storage API holds 48 of 64 in-use Postgres connections (75%)**, and the database has been logging `53300 remaining connection slots are reserved for roles with the SUPERUSER attribute` continuously for days.

PostgREST — which is what one would normally suspect — accounts for only 5 connections.

## Environment

- `hot-updater`, `@hot-updater/supabase`, `@hot-updater/server`, `@hot-updater/react-native`, `@hot-updater/core`: `0.31.4`
- Deployment: Supabase Edge Function (the template generated by `hot-updater init`)
- Database plugin: `supabaseDatabase`
- Storage plugin: `supabaseStorage`
- `updateStrategy: 'appVersion'`
- The client app itself never touches this database directly: its backend is Firebase, and it has no `supabase-js` / `pg` / Prisma / Drizzle dependency. Its only reference to Supabase is the Edge Function URL passed to `HotUpdater.wrap`. This project was created by `hot-updater init` and is used for OTA.

Code references below point at `main` (`0.35.x`), where the relevant code is unchanged.

## Symptom

Postgres logs, ~230 occurrences between 2026-08-08 and 2026-08-12:

```
level: error   status: 53300
remaining connection slots are reserved for roles with the SUPERUSER attribute
```

Two characteristic shapes:

- **Pairs 20–40 ms apart** — e.g. `12:43:14.454` / `12:43:14.478`, `10:28:08.831` / `.861`
- **Dense bursts** — e.g. 15 errors between `09:23:37.59` and `09:23:59.34`

Errors track daytime traffic and go quiet overnight, i.e. they follow app usage rather than a constant background load.

## Evidence

```sql
select usename, application_name, state, count(*)
from pg_stat_activity group by 1,2,3 order by 4 desc;
```

| usename | application_name | state | count |
|---|---|---|---:|
[...]

PR #1147, opened 2026-08-13
fix(supabase): batch signed URL generation

gronxb:
## Summary

- coalesce concurrent Supabase signed URL requests within the same event-loop turn
- issue one `createSignedUrls()` call per bucket instead of one `createSignedUrl()` call per changed asset
- preserve the existing runtime storage interface and map per-object failures back to the matching request
- apply the behavior to both generated Edge Function storage and universal Supabase storage

This keeps buckets private while removing the per-asset Storage API and Postgres lookup fan-out during update checks.

 #1145

## Verification

- `pnpm -w build`
- `pnpm -w lint`
- `pnpm -w test` (159 files, 2,098 tests)
- `pnpm --filter @hot-updater/supabase test:type`

### DIFF
--- a/.changeset/batch-supabase-signed-urls.md
+++ b/.changeset/batch-supabase-signed-urls.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/supabase": patch
+---
+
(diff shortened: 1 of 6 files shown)


gronxb:
@codex review

PR #1148, opened 2026-08-13
fix(storage): harden prune compatibility and guidance

gronxb:
## Summary

- clarify `storage prune` as reference-aware garbage collection with explicit dry-run output and a `--protect-newer-than` safety barrier
- page standalone prune reads within the server limit and chunk multi-ID standalone deletion lookups at 100 records
- preserve existing UUIDv7-named S3 channels without restoring per-UUID recursive LIST requests
- add actionable CLI help, a Storage Cleanup guide, AWS maintenance guidance, and a release changeset

## Issue #1131 outcome

The management slowdown came from unfiltered S3 discovery traversing every legacy root-level UUID bundle directory and repeatedly loading metadata during deletion. The scan fix from #1146 avoids that traversal and stores new bundle artifacts below `bundles/<bundle-id>`, so upgrading improves management scans without requiring storage deletion.

Existing UUIDv7 channel names remain compatible. When legacy UUID roots exist, the compatibility path performs a bounded, paginated scan only over the matching leading key range, excludes `assets/` and `bundles/`, and accepts only exact `<uuid-channel>/<platform>/<target>/update.json` keys. It never issues one recursive LIST per UUID directory.

Pruning is optional physical cleanup. The operator workflow is:

1. Select obsolete records with `hot-updater bundle list`, including `--target-app-version` when useful.
2. Delete the exact records with one multi-ID `hot-updater bundle delete <bundle-ids...> --yes` command.
3. Review physical cleanup with `hot-updater storage prune --dry-run`.
4. Stop deploy/promote writers and run `hot-updater storage prune --protect-newer-than 24h --yes`.

The protection duration uses storage object modification time, not time since record deletion. Separate databases or environments sharing a bucket must use separate storage `basePath` values before pruning.

## Follow-up to #1146 review

- standalone prune pagination stays at the server maximum of 100
- standalone multi-ID deletion chunks lookups at 100 and still commits once
- UUIDv7 S3 channels remain readable and writable while legacy artifact roots avoid per-prefix traversal

## Verification

- targeted regressions: 3 files, 114 tests
- built CLI `storage prune --help` verified with dry-run and deletion examples
- docs production build and dead-link checks passed
- full local FixCI sequence passed:
  - build: 26 projects
  - type-check: 34 projects
  - lint: 0 warnings, 0 errors
  - unit: 162 files, 2,129 tests
  - integration: 15 files, 1,454 tests

Follow-up to #1146 and #1131.

### DIFF
--- a/.changeset/bright-dodos-clean.md
+++ b/.changeset/bright-dodos-clean.md
@@ -1,0 +1,8 @@
+---
+"@hot-updater/aws": patch
+"hot-updater": patch
+---
+
+Preserve UUIDv7 S3 channels while avoiding per-prefix legacy traversal, respect
+standalone server pagination limits during storage pruning and batch deletion,
+and document the safe storage cleanup workflow.
--- a/.changeset/fix-s3-management-listing.md
+++ b/.changeset/fix-s3-management-listing.md
@@ -7,4 +7,4 @@
 "hot-updater": minor
 ---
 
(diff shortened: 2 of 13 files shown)


gronxb:
@codex review

PR #1170, opened 2026-08-20
fix(android): preserve Metro in debug builds

marqroldan:
## Summary

- Keep Metro as the JavaScript bundle source for Android debug builds.
- Use `HotUpdater.getJSBundleFile(...)` only for Android release builds.
- Cover both the RN 0.82+ `getDefaultReactHost` and RN 0.81/Expo host transformation paths.
- Add a patch changeset for `@hot-updater/react-native`.

## Why

The Android config-plugin transform currently injects `HotUpdater.getJSBundleFile(applicationContext)` unconditionally. In a debug build launched with `npm run android` / Expo, that bypasses Metro and can leave the app stuck on the launch screen instead of loading the development bundle.

The iOS transform is intentionally unchanged: its generated `#if DEBUG` branch already continues to use Metro.

## Validation

- `git diff --check`
- `pnpm dlx oxfmt@0.43.0 --check packages/react-native/plugin/src/transformers.ts packages/react-native/plugin/src/withHotUpdater.spec.ts`
- Focused Node 24 transformer assertions for both Android host patterns, idempotence, and the unchanged iOS debug path

The full workspace test suite was not run in the lightweight upstream checkout; CI should run the repository test matrix.

## Attribution

Generated by 5.6 Luna Max at the request of the Queue App team.

### DIFF
--- a/.changeset/metal-debug-bundle.md
+++ b/.changeset/metal-debug-bundle.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+Preserve Metro as the JavaScript bundle source for Android debug builds while continuing to use Hot Updater bundles for release builds.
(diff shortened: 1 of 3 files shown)


gronxb:
@codex review

PR #1188, opened 2026-08-24
fix(react-native): reject incomplete bundle manifests

gronxb:
Follow-up to #1185.\n\n## Summary\n\n- validate the manifest bundle ID and every declared asset before install or launch\n- reject ambiguous platform bundle candidates while retaining legacy no-manifest fallback\n- validate cached nested bundle paths on Android and iOS\n- add native regression tests and TAR.GZ PAX writer coverage\n- run Android JVM unit tests in CI\n\n## Verification\n\n- Full local FixCI: build, type, lint, 2,153 unit tests, 1,455 integration tests\n- Android JVM tests: 40 passed\n- Swift tests: 22 passed

### DIFF
--- a/.changeset/calm-bundles-check.md
+++ b/.changeset/calm-bundles-check.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
+
+Reject incomplete manifest-driven bundles before installing or launching them.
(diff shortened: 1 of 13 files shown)


gronxb:
@codex review

PR #1196, opened 2026-08-25
feat: select smaller OTA artifact delivery

gronxb:
## Summary

- Persist exact archive and patch byte sizes in the unreleased v1 Bundle contract across official database providers and generated schemas.
- Record the exact served representation size and hash for manifest assets, including content-addressed Brotli payloads, without changing the storage-provider contract.
- Plan update delivery on the server: count the manifest UTF-8 bytes plus each usable primary changed payload, prefer a patch only when it is strictly smaller than an available file, and return archive-only when the known diff cost is equal to or larger than the archive. Unknown or unusable metadata keeps the existing safe diff-first behavior.
- Keep the change next-only: no main compatibility layer, native protocol change, client HEAD request, or request-time storage metadata probe.

## Compatibility

This updates the initial, not-yet-released `1.0.0` database schema in place. Databases created from an earlier next build must be recreated; there is intentionally no incremental migration or backfill.

## E2E

Adds the shared `size-aware-artifact-selection` scenario for iOS and Android:

1. deploy a base and a small target, then prove the artifact response uses the manifest triplet plus changed assets;
2. deploy a large, highly compressible raw asset target, then prove the response is archive-only;
3. install each result and verify the stable UI state.

The scenario is stable-only because this PR changes server artifact selection, not native recovery behavior. It captures Release identity and Bundle artifact-response evidence separately.

## Verification

- `pnpm -w lint`
- `pnpm -w test:type` — 34/34 projects
- `pnpm -w test` — 260 files, 2311 tests
- `pnpm -w build` — 26/26 projects, docs dead links included
- `pnpm -w test:integration` — 25 files, 295 tests
- E2E TypeScript/JavaScript format and lint checks
- iOS and Android dry-runs for `size-aware-artifact-selection`
- `pnpm exec changeset status`


gronxb:
@codex review

PR #1197, opened 2026-08-25
fix(react-native): preserve nested iOS bundle paths

gronxb:
## Summary

- resolve cached bundle IDs relative to the bundle store root
- preserve the full nested bundle path when moving staged updates into the bundle store
- cover nested cache identity, manifest-driven installs, and moved bundle paths
- add a patch changeset for `@hot-updater/react-native`

Follow-up to #1188.

## Testing

- `swift test --package-path packages/react-native/ios/HotUpdater` (37 tests, 3 suites)

### DIFF
--- a/.changeset/fuzzy-dots-rest.md
+++ b/.changeset/fuzzy-dots-rest.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/react-native": patch
+---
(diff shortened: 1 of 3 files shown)


gronxb:
@codex review

PR #1205, opened 2026-08-25
feat(signing): simplify local and managed bundle signing

gronxb:
## Summary

- preserve the v0 built-in local signing API: `enabled: true` and `privateKeyPath`; `enabled: false` or omitting `signing` disables signing
- make local `publicKeyPath` optional; an explicit path pins the expected key and enables public-key-only native builds
- expose only `remoteSigning`, `awsKmsSigning`, and `googleCloudKmsSigning` from `hot-updater/signing`; these signing plugins still require `publicKeyPath`
- remove ambiguous signing factory names, signing-only internal subpaths, and provider-specific signers from AWS, Firebase, Cloudflare, and Supabase packages
- expose `createBundleSigningHandler` from the `@hot-updater/plugin-core` root for compatible signing services
- keep Console read-only and redact private-key paths and credentials

## Local compatibility

Existing v0 configuration continues to work without a separate public key file:

```ts
signing: {
  enabled: true,
  privateKeyPath: "./keys/private-key.pem",
}
```

- Local signing derives its public key from the configured private key and retains the existing RSA-SHA256 wire format.
- `publicKeyPath` is an optional pin for local signing. If explicitly supplied, missing, invalid, or mismatched keys fail rather than falling back.
- `enabled: false` requires no keys. A local config without `enabled: true` does not implicitly enable signing.
- `keys export-public` and `doctor` resolve the local private key, with a generated sibling public-file fallback for key inspection.
- Expo retains v0 `HOT_UPDATER_PRIVATE_KEY` (PEM or file path), private-file, and public-file resolution when no explicit public pin is configured. An explicit pin bypasses those private sources.
- Config loading does not read private keys. Console uses only an explicit or generated sibling public file; unavailable inspection does not prevent valid local signing.

## Provider model

- Local PEM remains the interoperable baseline, with key custody managed by the protected deploy environment.
- AWS KMS and Google Cloud KMS are the hardened custody options. Their optional SDKs are dynamically imported and not bundled into the CLI by default.
- Cloudflare and Supabase users can use local PEM, call AWS/GCP KMS directly from a protected deploy job, or connect a compatible service with `remoteSigning`.
- GitHub is not a signer. Actions can hold a protected local PEM secret or use OIDC to obtain short-lived AWS/GCP credentials.

## Security boundary

- local signing requires RSA keys of at least 2048 bits and accepts only the exact 32-byte bundle digest
- AWS fixes the signing algorithm and message type, validates response identity, resolves the provider public key, and verifies signatures locally
- Google Cloud pins a CryptoKeyVersion, validates response identity and CRC32C fields, and verifies signatures locally
- AWS custom endpoints require HTTPS except exact loopback development endpoints
- remote signing enforces authentication, fixed protocol/algorithm values, strict schemas and size limits, blocked redirects, and local signature verification
- native builds with an explicit public pin do not require signing credentials; legacy local key resolution remains available for compatibility
- signing authenticates bundle bytes only; release policy, access control, audit logging, rate limits, and rotation remain deployment responsibilities

## Documentation

- preserve v0 local examples and document optional public-key pinning separately
- document public-key-only native builds, legacy Expo/EAS compatibility, and Console's public-file-only boundary
- document AWS/GCP, GitHub Actions secrets and OIDC, Cloudflare/Supabase choices, least-privilege permissions, and native-first rotation
- update examples, managed-provider guides, plugin-system guidance, and the changeset
- distinguish preserved v0 local configuration from stricter RSA/SPKI checks, native-key mismatch validation, and non-overwriting key commands in `BREAKING_CHANGES.md`
- clarify that keeping the signing key does not remove the v1 native SDK rollout requirement
- document legacy ignore-rule migration for public-key-only builds without exposing private material
- keep Azure Key Vault, Managed HSM, Vault, and OpenBao outside this scope

## Verification

- build: 26 projects passed, including docs production build and dead-link checks
- typecheck: 34 projects passed
- lint: 0 warnings and 0 errors
- unit: 273 files, 2,442 tests passed
- integration: 25 files, 295 tests passed
- regressions cover v0 local signing without a public file, disabled config, explicit pin mismatch, native key export/doctor, Expo/EAS key sources, and Console redaction

### Final documentation audit (`9df4b25e9`)

Documentation-only follow-up; no runtime code changes.

- docs production build: 318 generated files; dead-link checks passed
- docs typecheck, workspace lint, and all 2,442 unit tests passed again
- 17 internal routes/anchors verified against generated HTML; four root migration links verified
- both GitHub workflow examples and eight TypeScript, one JSON, and eight Bash examples passed syntax checks


gronxb:
@codex review

PR #1210, opened 2026-08-29
fix(aws): apply Lambda@Edge memory size on every deploy

bigcupcoffee:
Context: I noticed our lambda downgraded back to 128MB, and it's constantly reaching that limit, getting slowed down by constant GC attempts. `createFunction` never passed `MemorySize`, so when I upgraded from 0.29 it created new lambda and it used 128MB by default

Claude:
That branch also published before it reconfigured: `updateFunctionCode` ran with `Publish: true` and `updateFunctionConfiguration` came after. A published version is an immutable snapshot of the configuration at publish time and `updateFunctionConfiguration` only mutates `$LATEST`, so the version handed to CloudFront always carried the previous deploy's memory size. Combined, 256 MB took three deploys to reach traffic

### DIFF
--- a/.changeset/lambda-edge-memory-size.md
+++ b/.changeset/lambda-edge-memory-size.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/aws": patch
+---
+
+Apply the Lambda@Edge memory size on every deploy: `createFunction` now sets `MemorySize`, and updates publish a version only after the configuration change is applied.
--- a/plugins/aws/iac/lambdaAsset.ts
+++ b/plugins/aws/iac/lambdaAsset.ts
@@ -1,0 +1,9 @@
+import path from "path";
+
+/**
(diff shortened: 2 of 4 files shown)


bigcupcoffee:
More context on performance as it was affecting us directly, TLDR:

Lambda became slow and very often went over 5s default timeout on react-native client. This is when I noticed 128MB memory limit again. Bumping memory back to 256 by hand improved cold p50 from ~2632 ms at 128 MB to 1594 ms (~39% better)

In the end I raised the timeout to 10s anyway, I'm not sure if 0.36 improved the scans on lambda but I'll see next week when I upgrade I guess

gronxb:
@codex review

gronxb:
Thanks for the contribution! I’m aware of the performance issues, and the upcoming 1.0 release should fundamentally improve performance. Thanks again!

PR #1293, opened 2026-09-09
refactor: simplify Insights storage and stabilize native recovery

gronxb:
Insights exposes canonical events while each provider maintains the minimum internal state needed for efficient current-installation queries. This implements the storage redesign on top of #1289 (`codex/insights-download-pending`) and corrects the history-dependent reads found during provider verification.

- Remove shared `bundle_installations` storage and the installation-row contract. SQL and MongoDB maintain a private, nine-field `bundle_event_heads` access index with at most one row/document per installation. Latest pages filter and limit heads before fetching canonical payloads; counts read heads without event history. Supabase records events and heads through one atomic RPC. DynamoDB adds compact six-field scope entries in its existing table; Firestore retains its existing latest-event documents and native counts.
- Simplify writes to `recordEvent({ event })`. `findLatestEvents` returns canonical events; core derives the running/pending response. `countLatestEvents` receives fixed OR predicates and counts each installation once. Plugin authors implement the specification without lifecycle helpers, reducers, or new background services.
- Reuse `metadata` for `username`, `cohort`, `update_strategy`, `fingerprint_hash`, and `sdk_version`. Keep query identities and transitions in columns, preserve SDK HTTP requests and Console responses, and validate known metadata fields at the core boundary.
- Include the PRD, SQL/NoSQL plugin-author personas and adversarial agreement, reference examples, and storage measurements. PRDs remain repository-only and are excluded from public pages, search, and generated LLM documentation.
- Align Firebase Functions/Admin/CLI versions, preserve JSON bytes through the Functions entrypoint, and grant the AWS runtime the release-catalog lookup permissions it uses. Firebase retains eight required composite indexes and excludes metadata from indexing; AWS reuses its existing GSI without adding another.

This is a breaking custom-provider contract and an unreleased schema change. Each provider keeps its single `1.0.0` initialization migration. One-off cleanup of development databases is excluded from the implementation: no RC conversion script, conversion tests, compatibility alias, extra migration, or public conversion procedure is shipped.

The earlier event-only SQL implementation made latest-state reads depend on retained history. This was the wrong trade-off for current-installation queries. The correction leaves the canonical-event plugin contract unchanged and keeps derived storage private. Event/head writes are atomic and advance by `(received_at_ms, id)`; duplicate, delayed, user/scope movement, and concurrent receipts preserve canonical data and global-latest-before-filter semantics.

Provider choices are checked with real execution plans and integration tests. SQL/MongoDB keep only user, scope, and from/to bundle head indexes and remove the obsolete event-history user index. MySQL uses two disjoint counts in one statement to avoid a broad OR range scan; PostgreSQL, SQLite, and D1 retain the indexed OR form. No extra DynamoDB GSI or Firestore index is added.

Local workerd D1 measurements use identical data and the checked current initialization schema:

| Current-state query | Event-only | Optimized |
| --- | ---: | ---: |
| Count 1,000 installations, 10 events each | 63,999 reads | 1,000 reads |
| Count 1,000 installations, 100 events each | 186,000 reads | 1,000 reads |
| User page of 101 events, 100 events per installation | 20,181 reads | 202 reads |
| Rare bundle count, 100 events per installation | 177 reads | 8 reads |

The existing E2E D1 database was also compared before and after the infrastructure change with the same cutoff and identical results. With all 2,422 canonical events preserved, the 114-installation scope count changed from 4,473 to 115 reads, and the same 50-result user page changed from 1,333 to 100 reads. Together these two queries read 5,806 → 215 rows (96.3% fewer). All 495 heads matched the globally latest canonical event. This is a bounded comparison of those queries, not the multiplier for account-wide daily usage.

A winning event for an existing installation changes D1 `rows_written` from 8 to 12, including index writes; a duplicate writes 0 and a delayed event writes 7. This is one canonical event insertion plus one head update, not 12 event records. These local engine measurements are not a production billing comparison or a causal multiplier for the quota incident. The decision record includes fixtures, schema hash, alternatives, storage costs, and MongoDB/DynamoDB/MySQL trade-offs.

Native recovery changes address the white screen found during provider verification. Controlled scheduling on the original Release binary proved that a Fabric start from the failed runtime could run against the recovery runtime's reused surface before its bundle loaded. Native recovery metadata alone therefore did not prove a usable recovery.

- iOS suppresses starts and readiness reports from the failed JS runtime. Starts queued before the fatal error finish their animation-driver setup before recovery reloads the host. Coordination uses completion notifications without blocking the main or JS thread; the hooks preserve React Native's original implementations.
- iOS preserves previous/default fatal handlers when recovery is unavailable, forwards a failed reload to normal error handling, and prevents stale content/monitoring callbacks from verifying a failed launch. Uncaught native exceptions preserve the crash marker for the next launch.
- Android serializes fatal/content decisions, lets the restart activity own process termination, and publishes complete crash markers before restarting. Required reflection fields remain available under R8.
- Android crash verification ends instrumentation before the initial crash launch, then observes the SDK's automatic restart, foreground process, and recovery receipt before reattaching Detox and checking visible UI. Android's instrumentation cleanup would otherwise stop the entire package when the SDK replaces its process, including the restart activity. Actual crash launches are explicitly marked in the scenarios. iOS crash launches disable Detox synchronization while the failed runtime is destroyed; normal launches restore default synchronization. The example's Release build enables minification and preserves the specific shared APIs referenced by the separate Detox test APK.

These changes are confined to HotUpdater and its verification code; no React Native dependency patch is included. Recovery retains the existing new-bundle startup boundary. A native networking exception seen in an earlier Firebase run has not been proven to share the Fabric race; passing reruns alone are not treated as its root-cause fix.

Validation (verification commit `d0ab9fc1b11e800c922364b2669d165c26c4a0ca`; final provider matrix complete):

- Full ordered local CI passes after the read-cost and IAM corrections: build, type check, lint, 2,732 unit tests, and 383 integration tests using Java 21.
- An independent R8 shrink/reflection probe without Detox reproduces each missing accessor and passes with the consumer rules. The E2E example includes Detox keep rules, so its minified pass alone is not the consumer R8 proof.
- A rebuilt minified Android app/test APK passes Detox attachment, Ready UI, and native view-attribute checks, with no ANR, fatal, or linkage errors. The example fingerprint script uses the same CLI entry point as E2E; pnpm no longer injects undeclared Expo autolinking sources through its executable shim.
- Native regression tests pass: 37 Swift storage/archive tests plus 3 recovery-coordination tests, and 31 Android JVM tests covering recovery arbitration, complete crash markers, and storage.
- Actual iOS simulator Release build passes. Both platform fingerprints were regenerated and match. Cloudflare, Firebase, Supabase, AWS, and all five standalone profiles each passed the full iOS + Android suite sequentially on the final commit: 26 unique scenarios per platform and provider, 234 per platform, 468 total. Earlier-commit results are not substituted for this matrix.
- All four managed example deployments now run the final implementation and pass real download/apply/recovery/unchanged receipts, running/pending state, user/scope movement, expected authentication responses, bounded Storage reads, and signed artifact downloads. Cloudflare and Supabase initialized 495 and 414 existing heads; AWS conditionally initialized 477 scope items without modifying canonical history or adding a GSI. Firebase preserves its 21 existing indexes, including eight required Insights composites. SQL migration history, endpoints, keys, and unrelated settings are preserved.
- Infrastructure doctor checks pass. Full doctor on the original example still reports its existing Android public-key/iOS signer mismatch; this is not counted as a green native check. The actual profile setup restores its persistent signing key and exports the public key before both native builds; the final full matrix validates that prepared app.

The full suite covers visible recovery, restored release/channel, rejecting a republished crashed bundle, and applying the next safe update. The ten-entry crash-history scenario seeds ten entries; it does not simulate ten actual crashes.

Final full provider matrix:

| Profile | iOS / Android | Native audit |
| --- | --- | --- |
| Firebase | 26/26 · 26/26 | Passed |
| Supabase | 26/26 · 26/26 | Passed |
| Cloudflare | 26/26 · 26/26 | Passed |
| AWS | 26/26 · 26/26 | Passed |
| standalone-dynamodb | 26/26 · 26/26 | Passed |
| standalone-drizzle | 26/26 · 26/26 | Passed |
| standalone-prisma | 26/26 · 26/26 | Passed |
| standalone-kysely | 26/26 · 26/26 | Passed |
| standalone-mongodb | 26/26 · 26/26 | Passed |

The final cumulative audit rechecked all nine jobs against the same commit and the complete scenario manifest. It verified 36 expected iOS crash recoveries, 36 automatic Android recoveries without instrumentation driving the restart, 432 Android lifecycle assertions, complete native log coverage, and zero ANR warnings or unexpected app-native failures. All 18 Android raw-log hashes still match the immutable reviews. Firebase's earliest review did not record source-log hashes; those two source logs were freshly parsed and hashed, while the other 16 source hashes matched their recorded values.

The Android audit checks SDK-owned termination before the new process starts and observes the old process death before restored content. An AWS audit alarm exposed that asynchronous [AMS death notifications](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main/services/core/java/com/android/server/am/ActivityManagerService.java#1518) can follow the new-process start log; the private audit now accepts both notification orders. Two independent reviewers and 15 captured-log controls verified the correction, including rejection of missing/wrong-owner kills, premature starts, missing death evidence, instrumentation interference, fatal/linkage errors, and mismatched clocks. Earlier profile results were revalidated without changing app source, assertions, or raw logs.

AWS first final-head attempt stopped before scenarios because the Android build JVM exhausted its 512 MiB Metaspace. The rerun passed the native build and all 52 scenarios with a temporary host-only 1024 MiB Metaspace limit (heap unchanged at 2048 MiB), without source or test assertion changes. After all nine profiles completed, the unchanged temporary override was removed and the original host configuration restored. Verification collectors were stopped; raw logs and independent reviews were retained.

The Console appearance from #1289 is preserved (Active / Downloaded / Recovered totals above chart tabs):

Console Insights


gronxb:
@codex review

PR #1295, opened 2026-09-09
chore: remove unused `aws-lambda` dependency

SirPole:
Hello!

While auditing dependencies in our codebase, where we use hot-updater, I noticed, that we pull the deprecated `aws-sdk` v2. Tracked the cause to `aws-lambda`, which is used as a dependency of `@hot-updater/aws` package. Hot-updater by itself uses aws sdk v3. I inspected the code and found out it is only using the types to type expected handler signature, so the `@types/aws-lambda` is sufficient to function properly and `aws-lambda` can be safely removed, as it is completely unused.

I've ran build, lint and tests, all of which have passed succesfully.

I've also ran `pnpm install` to update the lockfile. This appears to have changed some unrelated dependency resolutions, please review those changes as well.

### DIFF
--- a/.changeset/wide-taxis-sit.md
+++ b/.changeset/wide-taxis-sit.md
@@ -1,0 +1,5 @@
+---
+"@hot-updater/aws": patch
+---
+
+Remove unused `aws-lambda` dependency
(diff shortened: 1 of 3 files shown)


gronxb:
@codex review

gronxb:
Thanks !

SirPole:
Thank you! 🙌

gronxb:
Released in v0.36.12

PR #1301, opened 2026-09-12
fix(cli): read default iOS target app version from project.pbxproj

yungblud:
`getDefaultTargetAppVersion` reads the iOS version from `Info.plist` only:

```ts
case "ios":
  version = await getIOSVersion({ parser: "info-plist" });
```

The React Native template ships `Info.plist` with

```xml
<key>CFBundleShortVersionString</key>
<string>$(MARKETING_VERSION)</string>
```

`getIOSVersionFromInfoPlist` rejects any value containing `$(` — correctly, since it cannot resolve Xcode build settings — so it returns `null`, and `getDefaultTargetAppVersion` returns `null` with it. The actual version only lives in `project.pbxproj` as `MARKETING_VERSION`, which `getIOSVersionFromXcodeProject` already knows how to read.

Two visible consequences in `deploy`, for any iOS project left on the template's `Info.plist`:

- `deploy -i` prefills the "Target app version" prompt with the hardcoded `1.0.0` fallback instead of the project's version, so anyone who accepts the default publishes the bundle against the wrong app version.
- `deploy` without `-t` and without `-i` has no value to fall back to and exits with "Target app version not found in native files".

`getNativeAppVersion` already handles this exact case:

```ts
// Info.plist first, project.pbxproj as a fallback.
return getIOSVersion({ parser: ["info-plist", "xcodeproj"] });
```

This PR gives `getDefaultTargetAppVersion` the same parser chain. The ordering and its rationale are preserved — `Info.plist` still wins when it resolves, and `project.pbxproj` is only opened when it does not, so the synchronous pbxproj parse stays off the common path.

Android is untouched: `build.gradle` stores `versionName` literally, so it never had this problem.

### Test plan

`packages/hot-updater/src/utils/version/getDefaultTargetAppVersion.spec.ts` (new), covering:

- `$(MARKETING_VERSION)` in `Info.plist` → falls back to `MARKETING_VERSION` from `project.pbxproj`
- the same path with a two-segment version → `2.15` normalizes to `2.15.x`
- a resolvable `Info.plist` → wins, and `XcodeProject.open` is never called
- neither source resolving → `null`

### Note on the import change

`getDefaultTargetAppVersion.ts` imported its two siblings through the `@/` alias. Vitest's project configs do not pick up the root `resolve.tsconfigPaths`, so importing this module from a spec fails to resolve, and the module could not be unit tested at all. The two imports now point at the same files relatively, matching `getNativeAppVersion.ts` next to it. Happy to drop this and fix the alias in `vitest.workspace.mts` instead if you would rather keep the alias.

### DIFF
--- a/.changeset/ios-default-target-app-version-xcodeproj.md
+++ b/.changeset/ios-default-target-app-version-xcodeproj.md
@@ -1,0 +1,5 @@
+---
+"hot-updater": patch
+---
+
+Read the default iOS target app version from `project.pbxproj` when `Info.plist` holds an unresolved build setting. The React Native template ships `CFBundleShortVersionString` as `$(MARKETING_VERSION)`, which the plist parser cannot resolve, so `getDefaultTargetAppVersion` returned `null` for most iOS projects: `deploy -i` prefilled the "Target app version" prompt with `1.0.0` instead of the real version, and non-interactive `deploy` without `-t` failed outright. `getNativeAppVersion` already falls back to the same parser; this aligns the two.
--- a/packages/hot-updater/src/utils/version/getDefaultTargetAppVersion.spec.ts
+++ b/packages/hot-updater/src/utils/version/getDefaultTargetAppVersion.spec.ts
@@ -1,0 +1,119 @@
+import fs from "fs/promises";
+import path from "path";
+
+import { XcodeProject } from "@bacons/xcode";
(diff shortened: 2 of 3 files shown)


gronxb (approved):

gronxb:
Released in v0.36.16.

Thanks !

